PR #85, opened 2015-05-20
allow to use different configuration when calling application, to all…

jmecosta:
Hi,

ive created a plugin to support F# in sonar, and use FSharpLint as the main tool to verify coding rules. You can give it a try here https://github.com/jmecosta/sonar-fsharp-plugin/releases/tag/0.9

the relevant requet on Sonar mailing list is here: http://sonarqube.15.x6.nabble.com/fsharp-support-for-sonar-td5035544.html

Since i had the need to create a own console runner to gather sonar metrics, i run into this limitation that i could not overwrite the configuration. My solution for this is to include new methods into the runlint interface that contains the configuration. You can see my solution here: https://github.com/jmecosta/sonar-fsharp-plugin/blob/master/FsSonarRunner/FsSonarRunnerCore/InputConfigHandler.fs 

Now ive just included the application source into my project and applied the change there, but it would be nice to just use dll instead

Is it possible to consider this
thanks in advance

Jorge Costa

### DIFF
--- a/src/FSharpLint.Application/RunLint.fs
+++ b/src/FSharpLint.Application/RunLint.fs
@@ -186,8 +186,8 @@ module RunLint =
 
     let internal neverFinishEarly _ = false
         
-    /// Parses and runs the linter on a single file.
-    let parseFile pathToFile (errorReceived:System.Action<ErrorHandling.Error>) =
+    /// Parses and runs the linter on a single file and a configuration
+    let parseFileWithConfig pathToFile (configuration:FSharpLint.Framework.Configuration.Configuration) (errorReceived:System.Action<ErrorHandling.Error>) =
         let input = System.IO.File.ReadAllText(pathToFile)
 
         let lintInformation =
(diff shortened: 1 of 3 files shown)


duckmatt:
Hi, that looks great. 0.2 will include an update to the API to make it easier for other applications to include FSharpLint, work on the API is ongoing in a [separate branch](https://github.com/fsprojects/FSharpLint/pull/84). The option to pass in your own configuration is planned. 

Cleaning up the API is taking a little longer than expected, so it would be good to merge your changes into master. Currently the changes are failing to build, it looks like a reference to `System.Xml.Linq` needs to be added to the projects that reference `FSharpLint.Application`, would you be able to add those please?

`/Users/travis/build/fsprojects/FSharpLint/src/FSharpLint.Console/Program.fs(86,17): error FS0074: The type referenced through 'System.Xml.Linq.XElement' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Xml.Linq'.`

jmecosta:
Sure, but looks odd that this would brake those builds. In will push the
change in a bit.

I dont have any build failures perhaps because of vs2015.

jmecosta:
build is passing now, some tests are failing seems some issues not related with this changes. can you confirm

jmecosta:
ive retry and it passed.

duckmatt:
great, thank you

jmecosta:
Thank you. If you happen to use sonar feel free to comment on the quality
debt model that ive setup as default

duckmatt:
I'll try and give it a go tonight, I'll let you know how i get on :)

jmecosta:
just let you know ive manage to make the copy past detection work. ive updated the jar. last thing to do is to be able to configure fxcop to report issues on f# code. But perhaps you are not that interested on that

PR #86, opened 2015-05-20
add nuget package for api only

jmecosta:
suppose this is also useful for you.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -18,10 +18,12 @@ open System
 // The name of the project 
 // (used by attributes in AssemblyInfo, name of a NuGet package and directory in 'src')
 let project = "FSharpLint"
+let projectApi = "FSharpLintApi"
 
 // Short summary of the project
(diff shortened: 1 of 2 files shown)


duckmatt:
thanks, this is useful

jmecosta:
No prob. Thanks also

duckmatt:
The api package is on nuget now here: http://www.nuget.org/packages/FSharpLint.Core/

The [api has changed a bit](https://github.com/fsprojects/FSharpLint/blob/master/src/FSharpLint.Application/Lint.fsi) and included functions to deal with getting and loading configuration files

jmecosta:
Ok thanks for the notice, I try to upgrade as soon as I have a bit of free
time. But anyway the overloaded functions to run lint are still there?

jmecosta:
Ok I see that it did, will take look

jmecosta:
works great, thanks.

Issue #11, opened 2014-04-01
Add more rules to XmlDocumentation analyser


duckmatt:
https://github.com/duckmatt/FSharpLint/blob/master/src/FSharpLint.Rules/XmlDocumentation.fs

Rules off the top of my head to be added:

Module comment
Function (not inside of a binding expression) comment
Member comment
Type comment
Enum case comment
Union case comment
Record field comment

jhamm:
I will work on this one. jhamm/FSharpLint 

* [ ] Module comment
* [ ] Function (not inside of a binding expression) comment
* [ ] Member comment
* [x] Type comment
* [ ] Enum case comment
* [ ] Union case comment
* [ ] Record field comment

duckmatt:
That would be great. 

When I've been writing new rules the easiest way I've found to do it is to add a test you'd expect to generate a lint warning, and another test with the same code fixed to not generate a lint warning. Then you can put a breakpoint on the visitor and run the test under debug to compare the difference in the ASTs to see what needs to be matched against.

Let me know if you have any problems/questions

jhamm:
Thanks! That's good advice.

duckmatt:
Thanks @jhamm

PR #87, opened 2015-05-21
Add more XmlDoc Rules

jhamm:
PR for #11.

The rules will apply in all Access cases; Public, Internal, Private. If we want to ignore one or more (maybe Private), it should not be too difficult.

The rules will trigger if there is XmlDoc that only contains whitespace.

* [x] Module comment
* [x] Function (not inside of a binding expression) comment
* [x] Member comment
  * [x] Method
  * [x] AutoProperty
  * [x] Let (not nested)
* [x] Type comment
* [x] Enum case comment
* [x] Union case comment
* [x] Record field comment

### DIFF
--- a/ConfigurationSchema.xsd
+++ b/ConfigurationSchema.xsd
@@ -223,6 +223,7 @@
             <xs:element name="UnionDefinitionHeader" maxOccurs="1" minOccurs="0" type="Enabled" />
             <xs:element name="RecordDefinitionHeader" maxOccurs="1" minOccurs="0" type="Enabled" />
             <xs:element name="AutoPropertyDefinitionHeader" maxOccurs="1" minOccurs="0" type="Enabled" />
+            <xs:element name="LetDefinitionHeader" maxOccurs="1" minOccurs="0" type="Enabled" />
           </xs:sequence>
         </xs:complexType>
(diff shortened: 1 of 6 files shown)


jhamm:
I took a look at the failing AppVeyor build.

It failed at
`Test Failure : FSharpLint.FunctionalTest.TestMSBuildTask+TestMSBuildTask.FunctionalTestMSBuildTask`
with the error
`FSC : error FS1135: Unexpected error creating debug information file '...\FSharpLint.FunctionalTest.TestedProject.pdb'`

[StackOverflow](http://stackoverflow.com/questions/6886790/) is indicating that the fix is to either clean the solution or delete the /obj/ and /bin/ folders.

The previous build from earlier today passed.
I was able to run the build.cmd and it completed with no issues.

I hope this is a heisenbug. If it is not, can you point me in the right direction to correct this?

duckmatt:
Thanks for this great work, looks perfect :+1: In regards to having the rules only applied to constructs with specific access I think that would be good to have as a config option

That failing test is a flaky acceptance test (sometimes has problems as it locks some files another acceptance test uses, looking to fix this soon), your changes wouldn't have caused this to fail

PR #94, opened 2015-06-05
Add markdown documentation for the XmlDocumentation rules

jhamm:
Now that we have the additional XmlDocumentation rules and can configure the rule based on access, we should probably tell people about it ;)

If you would like me to modify the wording or formatting, I will be more than happy to take care of it.


duckmatt:
Very true, I'll update the github pages with the new docs when I get home tonight and try to get a new package out this weekend with your latest changes in it

The wording and formatting looks good to me. Thanks :)

PR #107, opened 2015-07-16
Default value should be a static property

### LINKED ISSUE #105: Provide default record values for Lint API
We currently use:

```fsharp
Lint.lintParsedFile 
                    { FinishEarly = None
                      Configuration = None //Configuration.defaultConfiguration
                      ReceivedWarning = None }
                    { Ast = ast
                      Source = source
                      TypeCheckResults = checkResults.GetCheckResults() }
```

It would be great if `OptionalLintParameters` record type has a `Default` property (providing minimally useful config) so that we could just use `Lint.lintParsedFile OptionalLintParameters.Default ...`.

Any customization can be easily done with `{ OptionalLintParameters.Default with ... }`.

The same thing could be applied for `FSharpLint.Framework.Configuration.Configuration` record type. At a first glance, it's hard for clients to figure out what should be passed in.

dungpa:
Fix #105.


dungpa:
Sorry for the noise. It seems the repo doesn't have a `.gitattributes` file.

duckmatt:
Thanks, and for letting me know about the `.gitattributes` file, I'll add one

PR #116, opened 2015-07-29
test on Mono 4.0.2

sideeffffect:


sideeffffect:
the travis job timed-out when downloading packages. can we do something about it?

duckmatt:
Thanks, I assume nuget was down when that build was run, looks to be building now

PR #146, opened 2015-12-17
Do not suggest wildcard tuple replacement on object expressions

### LINKED ISSUE #145: Wrong suggestion to replace tuple wildcard on object expressions
17-12-2015 13-17-36 - tuple lint

If I follow the suggestion, the code is invalid. Further this suggestion only appears on object expressions. Normal implementation of the interface by the type `B` doesn't get this suggestion.

#### Code

```
type I =
    abstract member Two : bool * bool -> bool

type B() =
    let x =
        { new I with
            member __.Two(_, _) = false }

    interface I with
        member __.Two(_, _) = false
```

mexx:
fixes #145

### DIFF
--- a/src/FSharpLint.Rules/Binding.fs
+++ b/src/FSharpLint.Rules/Binding.fs
@@ -146,6 +146,7 @@ module Binding =
             | _ -> None
 
         match breadcrumbs with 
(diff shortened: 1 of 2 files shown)


duckmatt:
Great, thank you for the report and fix

Issue #148, opened 2015-12-30
Found tab character (`\t`) in file. for tabs in a literal string


smoothdeveloper:
IMHO this shouldn't trigger this lint warning:
```fsharp
let a = @"a	b"
let b = """
a	b
"""
```

(those are tabs but github replace with spaces)

duckmatt:
Thanks for the report, I agree that's a bug in that it reports a warning there

smoothdeveloper:
@duckmatt / @mexx great, thanks for quick turnaround!

PR #150, opened 2015-12-31
fix for #148

mexx:
I've started to work on a fix for #148.

Currently the typography rules, and no tab rule especially, are implemented with the `PlainTextVisitor`. This visitor do not have sufficient information about the AST to be able to decide whether it's a literal string or not.
Any pointers to how the implementation should be done are welcome.

### DIFF
--- a/src/FSharpLint.Application/Lint.fs
+++ b/src/FSharpLint.Application/Lint.fs
@@ -206,12 +206,14 @@ module Lint =
                         if parsedFileInfo.PlainText.Contains("SuppressMessage") then
                             Ast.getSuppressMessageAttributesFromAst parsedFileInfo.Ast
                         else []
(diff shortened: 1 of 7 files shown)


duckmatt:
Thanks, this will be really helpful :)

Currently the `PlainTextVisitor` is given a `PlainTextVisitorInfo` which has the field `SuppressedMessages` which is populated from the AST by `getSuppressMessageAttributesFromAst` in `visitPlainText`. We could do the same to get all the ranges of all strings, it's unfortunate in that it's an edge case, but I think that would be the simplest approach

mexx:
@duckmatt Please have a look, I've updated the PR.

duckmatt:
Thank you, that's great work

PR #194, opened 2016-12-01
Provide automated fix info

duckmatt:


Krzysztof-Cieslak:
I've found small bug:

```
[1;2;3]
|> List.map (fun x -> [x])
|> Array.concat
```

is suggested to:
```
List.collect fun x -> [x] [1;2;3]
```

parenthesis around lambda are missing

duckmatt:
@Krzysztof-Cieslak thanks for pointing that out, got a fix in for it. Currently some of the formatting is pretty dire, will make a separate release for those fixes as the code is at least syntactically correct e.g.

    let x y =
        y
        |> List.filter (fun x -> true)
        |> List.map id
        |> List.map id
        |> List.filter (fun x -> true)

turns into:

        let x y =
            (List.map (id >> id) (y 
            |> List.filter (fun x -> true)))
            |> List.filter (fun x -> true)

PR #207, opened 2017-01-21
Add a hint for Array.append

smoothdeveloper:
Array.append a (Array.append b c) ===> Array.concat [|a; b; c|]

See https://github.com/Microsoft/visualfsharp/pull/2296

Not sure it matters for List and Seq so I left those out, but nested Array.append are generating wasteful intermediary arrays, that can be avoided with Array.concat.


duckmatt:
Great, thank you 👍

PR #212, opened 2017-02-10
Add hint for x.average(x.map) to x.averageBy

ErikSchierboom:
This PR adds hints to convert `List.average (List.map x y)` into `List.averageBy x y`, and the equivalent for `Array` and `Seq`.


duckmatt:
Thank you

PR #213, opened 2017-02-11
Add hint for x.concat(x.map) to x.collect

ErikSchierboom:
This PR adds hints to convert `List.concat (List.map x y)` into `List.collect x y`, and the equivalent for `Array` and `Seq`.


ErikSchierboom:
Totally missed the existing hint, sorry about that!

PR #239, opened 2017-08-31
Update dependencies

duckmatt:


paket.lock
```diff
@@ -1,37 +1,38 @@
...
-    FSharp.Compiler.Service (13.0)
-      System.Collections.Immutable (>= 1.3.1)
+    FAKE.Lib (4.63)
+    FParsec (1.0.3)
+      FSharp.Core (>= 4.0.0.1)
+    FSharp.Compiler.Service (14.0.2)
+      System.Collections.Immutable (>= 1.2)
       System.Reflection.Metadata (>= 1.4.2)
-    FSharp.Compiler.Service.ProjectCracker (13.0)
-    FSharp.Core (4.0.0.1)
```
enricosada: can you leave fsharp.core `FSharp.Core (4.0.0.1)` if not really needed the upgrade?
so can be used by 4.0 tool, like FSAC

enricosada:
added a comment about fsharp.core version, so we dont need to upgrade in FSAC, if not needed here

paket.lock
duckmatt: yep, that's not a problem, i'll leave it as it at `4.0.0.1`

enricosada:
thx a lot @duckmatt !

duckmatt:
@enricosada No problem! :)

PR #245, opened 2017-11-27
Update template.cshtml - remove CyclomaticComplexity from menu

Krzysztof-Cieslak:
CyclomaticComplexity analyzer is no longer in FSharpLint, and clicking on the link results in 404


duckmatt:
Thank you

PR #253, opened 2018-02-19
Develop/issue 252

SteveGilham:
You may want to just cherry-pick the change https://github.com/fsprojects/FSharpLint/pull/253/commits/b4b2ad58056a97c8fcbf4f254f02b2b85d208b68, which has the substance.

https://github.com/fsprojects/FSharpLint/commit/c249471b6958e104d9b49695a15ed5aba35d27f3 covers files that were modified by running `build.cmd`.  The next two changes cleared the decks of compile-time warnings, plus one of those "helpful" things that Visual Studio does when opening a solution.


SteveGilham:
https://ci.appveyor.com/project/duckmatt/fsharplint-231/build/1.0.651 looks like it hit a race that https://ci.appveyor.com/project/duckmatt/fsharplint-231/build/1.0.650 avoided, the only change between the two being in the .net core solution file.

duckmatt:
Thank you for taking a look into this and putting the PR together. 

I've looked at the failing tests and I believe it's because `GetProjectOptionsFromScript(file, source)` is not returning any references other than `FSharp.Core` on dotnet core, there's an `assumeDotNetFramework` parameter which seems to give better results when true but is still missing `System.Runtime` and others.

SteveGilham:
About the unit tests -- I hadn't drilled into them, but I'm not surprised,
given quite how much is done by implication in the severely cut down new
project format.

Issue #256, opened 2018-02-21
MSBuild errors should cause a build to fail


SteveGilham:
### Description

Having switched on lint warnings as errors, and generated a screenful of red text, I expect a build failure; instead the build continues.

### Repro steps

I overrode the default target in an fsproj file by inserting
```
	<Target Name="FSharpLint" BeforeTargets="AfterBuild" Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
    <FSharpLintTask Project="$(MSBuildProjectFullPath)" TreatWarningsAsErrors="true"/>
	</Target>
```
after the NuGet introduced section for FSharpLint.

I then deleted the FSharpLint.settings file that had been suppressing some warning families, and ran a build to validate the behaviour when there were unsuppressed issues present.

### Expected behavior

The build should behave as if there were compilation errors.

### Actual behavior

The build continues as if there were no more than warnings emitted, because `FSharpLint.MSBuild.FSharpLintTask.Execute` returns `true` even after it has emitted errors, rather than the value `(not this.TreatWarningsAsErrors) || (Seq.isEmpty warnings)`.

### Known workarounds

My current work-around is to add a FileLogger to the MSBuild invocation that captures only errors, and then separately fail if the generated file is non-empty
```
msbuild mysolution.sln /flp1:ErrorsOnly; ...
```
### Related information 

* Windows 10 Home 1709
* Nuget package FSharpLint.MSBuild 0.9.0
* .net 4.7.x (Visual Studio v15.5.5)

PR #257, opened 2018-02-21
Addressing Issue 256

SteveGilham:
Signal that the build task has not completed with success should it have emitted errors.


duckmatt:
Great, thank you

fixes #256

PR #258, opened 2018-02-22
Issue 252

duckmatt:


src/FSharpLint.Core/Framework/ParseFile.fs
```diff
@@ -60,26 +61,55 @@ module ParseFile =
                   File = file } |> Success
             | Failed(_) -> Failed(AbortedTypeCheck)
         | None -> Failed(FailedToParseFile(parseResults.Errors))
+        
+    // See: https://github.com/fsharp/FSharp.Compiler.Service/issues/847.
+    let private dotnetCoreReferences () =
```
duckmatt: ideally we'll fix upstream but this should do the trick for now

PR #270, opened 2018-10-11
Add Formatting analyser

jgardella:
Adds a Formatting analyser with several rules from the [Microsoft F# formatting guide](https://docs.microsoft.com/en-us/dotnet/fsharp/style-guide/formatting). Specifically, the rules added are:

- `TypedItemSpacing` checks for spaces around colon in typed item
- `TupleCommaSpacing` checks for single space after comma in tuple
- `TupleParentheses` checks for parentheses around tuple instantiation
- `TypePrefixing` checks generic type prefixing rules
- `PatternMatchClausesOnNewLine` checks that every pattern match clause is on its own line
- `PatternMatchOrClausesOnNewLine` checks that every pattern match 'or' clause is on its own line
- `PatternMatchClauseIndentation` checks that pattern match clauses are correctly indented
- `PatternMatchExpressionIndentation` checks that pattern match expressions are correctly indented
- `ModuleDeclSpacing` checks that there are 2 spaces between module declarations
- `ClassMemberSpacing` checks that there is 1 space between class members
- `UnionDefinitionIndentation` checks for correct indentation in union definition

Currently none of these rules are configurable; they use configuration specified by the style guide. Configuration for these rules can be added in the future.


duckmatt:
Thanks for the great work

PR #272, opened 2018-10-11
Add indentation rule

jgardella:
Adds a configurable `Indentation` rule to the Typography analyser, which checks that indentation uses a consistent number of spaces (4 by default).

### DIFF
--- a/src/FSharpLint.Core/DefaultConfiguration.FSharpLint
+++ b/src/FSharpLint.Core/DefaultConfiguration.FSharpLint
@@ -327,6 +327,11 @@
           <OneSpaceAllowedAfterOperator>true</OneSpaceAllowedAfterOperator>
           <IgnoreBlankLines>true</IgnoreBlankLines>
         </TrailingWhitespaceOnLine>
+
+        <Indentation>
+          <Enabled>true</Enabled>
+          <NumberOfIndentationSpaces>4</NumberOfIndentationSpaces>
+        </Indentation>
       </Rules>
       <Enabled>true</Enabled>
     </Typography>
--- a/src/FSharpLint.Core/DefaultConfiguration.FSharpLint
+++ b/src/FSharpLint.Core/DefaultConfiguration.FSharpLint
@@ -329,7 +329,7 @@
         </TrailingWhitespaceOnLine>
 
         <Indentation>
-          <Enabled>true</Enabled>
+          <Enabled>false</Enabled>
           <NumberOfIndentationSpaces>4</NumberOfIndentationSpaces>
         </Indentation>
       </Rules>
--- a/tests/FSharpLint.Core.Tests/Rules/TestTypographyRules.fs
+++ b/tests/FSharpLint.Core.Tests/Rules/TestTypographyRules.fs
@@ -297,3 +297,26 @@ let pascalsTriangle =
     ]"""
 
         Assert.IsTrue(this.NoErrorsExist)
+
+    [<Test>]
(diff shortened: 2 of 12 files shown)


tests/FSharpLint.Core.Tests/Rules/TestTypographyRules.fs
```diff
@@ -207,4 +217,24 @@ type TestTypography() =
...
+        this.Parse """
+module P
+
+let x =
+  x"""
+
+        Assert.IsTrue(this.ErrorExistsAt(5, 0))
+
+    [<Test>]
+    member this.``No error for correct indentation``() =
```
duckmatt: Would also be worth adding a test case for places where the lines may not necessarily be indented as per the rest of the file e.g. below on line 3:
```f#
let rainbow =
    { Boss = "Jeffrey"
      Lackeys = ["Zippy"; "George"; "Bungle"] }
```

src/FSharpLint.Core/DefaultConfiguration.FSharpLint
```diff
@@ -327,6 +327,11 @@
           <OneSpaceAllowedAfterOperator>true</OneSpaceAllowedAfterOperator>
           <IgnoreBlankLines>true</IgnoreBlankLines>
         </TrailingWhitespaceOnLine>
+
+        <Indentation>
+          <Enabled>true</Enabled>
```
duckmatt: I think we should either have this disabled by default or try to auto detect their indentation, otherwise anyone with different valid indentation might have to start adding lint config files (or more likely disable the linter in their editor)

tests/FSharpLint.Core.Tests/Rules/TestTypographyRules.fs
jgardella: Yes this should be added. In fact, my current implementation will fail for this case. I'll add a test for this and try to fix it.

Do you know of anything else which would lead to different indentation like this?

src/FSharpLint.Core/DefaultConfiguration.FSharpLint
jgardella: Agreed.

tests/FSharpLint.Core.Tests/Rules/TestTypographyRules.fs
jgardella: I've updated the implementation and added some tests for this case.

duckmatt: Thanks for the addition of the tests, there's some good docs on indentation here: https://github.com/fsprojects/fantomas/blob/master/docs/FormattingConventions.md#general-rules-for-indentation these could be good test cases. Unfortunately looking at that there's a bunch of additional edge cases :(

jgardella: Thanks for the link Matt, very useful. I looked through it and added handling and test cases for all the exceptional indentation that I saw. Specifically:
- Record field indentation
- Array item indentation
- Pipelining
- Object expressions

There is also the exceptional case of aligning multiple usages of an infix operator, but I'm not sure how to implement this. Did I miss any other cases?

duckmatt: No problem. The changes look good, I can't think of any other cases off hand, I think it will be difficult to cover all cases up front, I'll enable it for my projects and see if I spot anything :)

PR #277, opened 2018-10-28
Move to dotnet core

duckmatt:


duckmatt:
@ErikSchierboom no problem! i'm very happy to leave behind the old project file formats :)

PR #279, opened 2018-11-08
Implement RecursiveAsyncFunction rule.

jgardella:
Checks for tail-recursive async functions which make their recursive call with `do!`. This can cause memory leaks; the recursive call should be made with `return!`. See [this stack overflow](https://stackoverflow.com/questions/33481220/memory-leak-in-f-async-recursive-call) question for more info.


duckmatt:
nice work!

PR #287, opened 2018-12-11
Fix guard indentation in FormattingMatchExpressionIndentation rule

### LINKED ISSUE #286: Unsure of correct styling for multiline when clauses in match expressions
### Description

The match expression with multi line when clauses can has unexpected desired behavior (see below)

### Repro steps
Consider the following block

```

namespace Placeholder

module testModule = 

    let testvar = match "x" with
                  | "a" when 5 > 3 &&
                             4 < 8 &&
                             2 > 9 -> "result"
                  | _ -> "otherresult"


```
Fsharp lint gives the following error

>If RHS of pattern match case is on newline, it should be indentend by one level.
>Error in file newline.fs on line 9 starting at column 38
>                             2 > 9 -> "result"
                                      ^  

I believe the following is the desired styling
```

namespace Placeholder

module testModule = 

    let testvar = match "x" with
                  | "a" when 5 > 3 &&
                             4 < 8 &&
                             2 > 9 ->
                      "result"
                  | _ -> "otherresult"

```

If this is the desired behavior then I suppose that's acceptable, I simply find this a bit surprising.

jgardella:
We were not taking the guard of a pattern match into account for the match expression indentation rule in the Formatting analyser.

Fixes #286.

### DIFF
--- a/src/FSharpLint.Core/Rules/Formatting.fs
+++ b/src/FSharpLint.Core/Rules/Formatting.fs
@@ -214,8 +214,12 @@ module Formatting =
             if isEnabled && isSuppressed ruleName |> not then
                 clauses
                 |> List.iter (fun clause ->
-                    let (SynMatchClause.Clause (pat, _, expr, _, _)) = clause
-                    if expr.Range.StartLine <> pat.Range.EndLine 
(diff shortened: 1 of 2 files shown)


duckmatt:
nice! thanks for the fix and sorry for the slow merge. i'll try to get a new version up

PR #300, opened 2019-01-29
Update default config to disable formatting rules for now

duckmatt:


duckmatt:
@jgardella I've updated the default config for the formatting rules so that for now they have to be explicitly enabled. I'm thinking it might be worth us running the linter on some of the larger well known f# projects to identify rules that maybe quite opinionated within the community, and we can start enabling by default the ones that are clear cut and are rarely broken

Went through a similar process a while back and made the more style based rules other than naming not enabled by default. Went with the goal of trying to have very few people require adding config to disable rules, found that people are much happier to add a file to enable a rule than add a file to disable a rule

jgardella:
Yes, that makes sense. I had set them as enabled by default because they were part of the Microsoft style guide but I guess that guide isn't as established as I thought. There are some issues to fix with the new rules so this makes sense in that context too.

Issue #313, opened 2019-02-13
Bump FCS to 26.0


Krzysztof-Cieslak:
We're slowly working on updating FSAC/Ionide to FCS 26.0 and we'll need also update FSharpLint to it. 

Details: https://github.com/fsharp/FsAutoComplete/pull/338

@baronfel

PR #314, opened 2019-02-13
Update fcs

baronfel:
This PR updates FSharpLint to use FCS 26 and FSharp.Core 4.5.2, as required by FCS 26.  This update introduced new F# 4.5 and 4.6 members, namely anonymous records, and several API changes to the AST:

* StructTuples are now signaled by the first `isStruct: bool` property of the `Tuple` DU case.  Wherever FSharpLint used struct tuples explicitly there's now a match for `Tuple(true, ....)`
* Match expressions no longer have the `isExnMatch: bool` property, so this was removed where bound
* Some type hints were added where methods have new overloads due to me running on a more recent mono. This fixes some type resolution errors, and is fully backwards-compatible.
* SynModuleOrNamespace no longer has a `isModule: bool` property, now it has a DU signaling if it is a `NamedModule | AnonModule | GlobalNamespace | DeclaredNamespace`. Where code was previously relying on `isModule` being true, I replaced it with a match for `(NamedModule | AnonModule)`.

That should cover the major changes. All tests pass on my machine.


baronfel:
This will close #313.

duckmatt:
Great, thank you! Changes look good and thanks for describing the changes made. 0.10.5 should be up on nuget in the next few minutes

baronfel:
Glad you approve, you'll be getting another one of these shortly after the next FCS merge :D

PR #322, opened 2019-02-27
improve build

enricosada:
- pin .NET Core Sdk version to v2.1.500 to make build deterministic
- use paket `storage: none` to speedup restore (less I/O)
- consider only used target frameworks to minimize package graph


duckmatt:
awesome, thank you!

PR #330, opened 2019-04-29
Redesign & refactor linter

jgardella:
This is a large PR to rewrite and refactor FSharpLint so that the code is clearer and more easily testable. Changes are listed at a high level below:

- Remove concept of analysers. Analysers previously served as way to group rules together. I think the one useful part of this was to allow disabling whole groups of rules, but I think the groupings were erratic enough for this to not be super useful.
- Each rule now lives in its own file. Each exposes a value `rule` which is either the rule itself, or a function from the config for the rule to the rule. Previously, config was loaded from within a rule's definition. Config loading is now done completely before rules are run, and then passed into the rules here on creation. This nicely separates rule configuration from rule logic.
- Two basic types of rules are provided:
  - `AstNodeRule` takes an `AstNode` as input (most rules are this type)
  - `LineRule` takes a line of source as input (use for typography analysis)
- Update the linter entrypoint to work with these new rules. This should help with performance. Previously, we had to loop through the syntax array of AST nodes for every analyser. Now, we only loop through twice; once for the `AstNodeRule`s and once for a couple of rules which require a context to be built up based on the syntax array (just `Indentation` and `NoTabCharacters` currently).
- Adds unique identifier of format `FS0001` to every rule. This prefixes any errors to allow for easy identification.
- Config is now specified as JSON instead of XML. Added a target to the dotnet tool to convert from XML to JSON config: `dotnet fsharplint -convert <xml_config_path> <output_path>`
- Update all tests for these changes. Tests for each rule now live in their own file, rather than being grouped by analyser.
- Updated all docs to reflect these changes.


docs-gen/markdown/index.md
```diff
@@ -82,46 +85,45 @@ The tool also used to have MSBuild and Fake tasks to run as part of a build. The
...
-	[lang=xml]
-    <?xml version="1.0" encoding="utf-8"?>
-    <FSharpLintSettings>
-      <Analysers>
-        <FunctionReimplementation>
-          <Enabled>False</Enabled>
-        </FunctionReimplementation>
-      </Analysers>
-    </FSharpLintSettings>
+    dotnet fsharplint -convert <path-to-xml> (output-path>
```
duckmatt: small typo on `(output-path>`

duckmatt:
Very nice, the breaking apart of the analysers into the individual rules has made things much neater! The linter's tests have good coverage of the behaviour so I'm confident all the logic has remained after the refactoring

PR #342, opened 2019-05-27
Update to FCS 29.0

Krzysztof-Cieslak:
One breaking change in the API - FSharpChecker is now using `ISourceText` instead of `string` for passing content of the file.


duckmatt:
thank you, will get a new version up

Krzysztof-Cieslak:
Cheers

duckmatt:
No problem, should be up as `0.12.0` shortly. 
It's worth noting the linter since `0.11.0` expects config files to be json and xml config is no longer supported, but the fsharplint dotnet tool has a converter to get a json file from the xml

PR #354, opened 2019-06-01
convert XmlConfiguration to new Configuration

milbrandt:
refactor to extentd public API allowing applications to migrate their old style XmlConfiguration to new ConfigurationManager.Configuration.
This allows consumers relying relying on Analysers in their application to migrate gradually to the new style.


jgardella:
Thank you @milbrandt!

PR #359, opened 2019-06-30
Update FCS to 30.0.0

baronfel:
Hi! There's a new FCS and so here's the update, along with a few changes I had to make to adjust to the new APIs.

The major changes are:
* SynAttributes isn't just a list anymore, so I made an extractor function to pull out the nested lists so surrounding code in FSharpLint didn't have to change
* SynSimplePats isn't just a list anymore (and the inner DUs are different) so again I made an extractor function to pull out the nested SynSimplePat so surrounding code in FSharpLint didn't have to change
* after the change, the `getUnionCaseStartColumn` function was causing two tests to error. I adjusted the code after some debugging and printfing of the relevant ranges to just use the StartColumn of the first attribute attached to the union case, if one was present. This range seems to contain the `[<` starter-sigil, and so this is the 'first column' of the overall DU case definition, and the one that should both be indented from the parent type definition and at the same depth as any sibling DU cases.  As a result, that function no longer needed to extract out the text at a range, so I removed the input source text parameter and adjusted callsites.


jgardella:
Thank you @baronfel!

baronfel:
Anytime! I'm going to try and continue doing this pattern to make FCS upgrades less impactful on our downstream consumers.

PR #369, opened 2019-10-01
Update to FCS 32.0

Krzysztof-Cieslak:


duckmatt:
Thank you, will publish a new version

PR #378, opened 2020-02-06
Update to FCS 34.0.1

baronfel:
This was pretty same-y, with two notable exceptions:

* I updated the reference-generation logic for the tests by specifically asking for .net core/.net framework-style references
* FCS 34.0.1 brought support for `and!`, but I don't know of any test examples that would be easy to add. I'd love to get some samples in and get this under test before it's merged here.


src/FSharpLint.Core/Application/Lint.fs
```diff
@@ -49,7 +49,7 @@ module ConfigurationManagement =
     /// The closer they are to the project directory the higher precedence they have.
     /// e.g. if the project directory is C:\User\Matt\Project then a config file found in
     /// C:\User\ will be loaded before and overridden by a config file found in C:\User\Matt\.
-    let private loadUserConfigFiles projectFilePath defaultConfig =
+    let private loadUserConfigFiles (projectFilePath: string) defaultConfig =
```
baronfel: these annotations were necessary to 'fix' new overloads

src/FSharpLint.Core/Application/Lint.fs
```diff
@@ -375,7 +375,7 @@ module Lint =
 
         exitCode, (workingDir, exePath, args)
 
-    let getProjectFileInfo (releaseConfig : string option) projectFilePath = 
+    let getProjectFileInfo (releaseConfig : string option) (projectFilePath: string) =
```
baronfel: these annotations were necessary to 'fix' new overloads

src/FSharpLint.Core/Application/XmlConfiguration.fs
```diff
@@ -337,7 +337,7 @@ module Configuration =
     /// The closer they are to the project directory the higher precedence they have.
     /// e.g. if the project directory is C:\User\Matt\Project then a config file found in
     /// C:\User\ will be loaded before and overridden by a config file found in C:\User\Matt\.
-    let tryLoadUserConfigFiles projectFilePath =
+    let tryLoadUserConfigFiles (projectFilePath: string) =
```
baronfel: these annotations were necessary to 'fix' new overloads

src/FSharpLint.Core/Framework/Ast.fs
```diff
@@ -386,7 +386,15 @@ module Ast =
             add <| expressionNode expression2
             add <| expressionNode expression1
             add <| expressionNode expression
-        | SynExpr.LetOrUseBang(_, _, _, pattern, expression, expression1, _)
+        | SynExpr.LetOrUseBang(_, _, _, pattern, rightHandSide, andBangs, leftHandSide, _) ->
+            add <| expressionNode rightHandSide
+            add <| expressionNode leftHandSide
+            // TODO: is the the correct way to handle the new `and!` syntax?
+            andBangs |> List.iter (fun (_, _, _, pattern, body, _) ->
```
baronfel: the and! syntaxes are of type `list<SequencePointInfoForBinding * bool * bool * SynPat * SynExpr * range>` here, so the interesting things are the pattern and the body, which I've extracted here.

I'm not sure this is correct, though.

baronfel:
Ok, this is green now @duckmatt. Do you have any thoughts or concerns about my changes here?

src/FSharpLint.Core/Framework/Ast.fs
duckmatt: Sorry for the slow response on this, I've had a quick look this morning before work and it looks good, but I do notice all the others are added in reverse order. Have to admit I can't actually recall why, I'll take a look after work and find out how important it is for them to be in order, and if so get a test to cover it

PR #379, opened 2020-02-10
Added a sample on how to inhibit the FS0050 warning

MecuSorin:
Hi! I added a sample on how to inhibit a naming warning on the page that will most likely will be hit by the search engines.

**NOTE: Didn't build the solution (so the html files were not regenerated) because I don't have installed the required SDK(switching the global.json to 3.1 didn't help).**

 If you consider that this will help, please do the required step :)

Cheers


jgardella:
Thanks a bunch for the PR @MecuSorin. I recently rewrote suppressions to use structured comments instead of attributes, so this change is no longer necessary and I will close it. However, please do take a look at the updated docs and see if you think anything is missing.

PR #384, opened 2020-02-24
Remove nested config loading

jgardella:
This PR contains several changes:

- **Removes logic for nested config loading.** I think the linter should support only a single config, with a default path in the root of the repository which can be overridden. I think the ability to override the config per project is not very useful, and rules can be disabled per-file if needed using suppression.
- **Remove XML configuration.** Code for the old XML configuration is removed, and the config conversion logic is also removed. I think users should be converted to the new JSON config by now.
- Add `--lint-config` param to dotnet tool to allow overriding the default path to the config (`fsharplint.json`).


jgardella:
@duckmatt Would like your feedback here when you get a chance, since this is a pretty fundamental change to how configuration works in the linter. I like this change as I was able to remove a lot of code, and I don't think the ability to have multiple configs which override each other is very useful; I would rather encourage users to have a single config per repo and then suppress rules where they need.

Also, is it reasonable to remove the XML configuration code at this point?

I am considering also doing a final change to the configuration to remove the grouping of rules by what used to be analysers (e.g. `formatting`, `naming`). This is just a hold over from when rules were grouped by analysers, and I think it would be clearer and less confusing to just have each rule defined at the root of the config. But I am a bit hesitant to introduce another breaking change to the configuration.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,6 @@
 ##### 0.13.1
 
+* Remove support for multiple overriding configs
```
duckmatt: Worth bumping the version for this with 0.13.1 already published

duckmatt:
I think this is a good change. To give a bit of a background on the original config overriding approach, this thought was taken from StyleCop, I'd found this genuinely useful in the past when working at a place that enforced it quite religiously (fail build on error etc). Rather than having multiple config files in a single repo the real benefit is having a config in your develop directory which then applies to all your repos which you can override for individual repos as needed. However I've never felt the need to do this with f# and the linter, plus having the lint rules explicitly in the project if requiring modification is a better approach

RELEASE_NOTES.md
jgardella: Whoops, I meant to release all these breaking changes together; I thought not putting the date would make it not trigger a release.

jgardella:
Thanks for taking a look. Regarding the breaking change for the config, sounds good. I agree about the v1 as well, I am trying to get the repo to a more stable state and fix some of the outstanding bugs, and then I think we can roll out a v1 release.

milbrandt:
The removal of XML config is breaking external tools still relying on XML like https://github.com/jmecsoftware/sonar-fsharp-plugin. IMHO this change was rather big change for a patch release.

duckmatt:
Apologies @milbrandt we're hoping to hit v1 soon, after that any breaking changes like this will be a bump in the major version

PR #388, opened 2020-02-27
Add new config version fields

jgardella:
Updates the config to remove grouping of rules by the old analysers they used to belong to. Each rule is now defined at the base level of the config object. Also updates each rule to use the correct Pascal C ased name to align with the rule's name and the documentation. This is an non-breaking change; the old grouped rule configs are marked as deprecated, and will be removed in the next major release.


jgardella:
@duckmatt Please take a look when you get a chance. I would still like to add a warning in the when deprecated config fields are detected.

duckmatt:
Nice, this looks good!

PR #389, opened 2020-02-27
Run linter against repo

jgardella:
Adds a targets file to trigger linting when building projects in the repo, and a config file.


jgardella:
@duckmatt In order to get the linting to work, I had to make all projects target only a single framework, since the call to `getProjectInfos` in `dotnet-projinfo` doesn't work for multi-targeted projects. Is it okay to only target `netstandard2.0` at this point, or do we still need to target `net461`?

duckmatt:
@jgardella Yep, should be fine to target just netstandard2.0 now

jgardella:
@duckmatt I agree it's not ideal; it's a limitation of `dotnet-projinfo`. I'm not sure if this is an inherent limitation in how MSBuild works or if the `dotnet-projinfo` could be updated to handle it. @enricosada would know more.

jgardella:
@duckmatt After removing the net461 target, the CI seems to be failing. Any idea why that would happen? Do we need to update something in the CI configurations?

duckmatt:
@jgardella had a quick look this morning, seems to be failing at `Dotnet.ProjInfo.Inspect.getProjectInfos`

I can get the build to succeed if I build the `FSharpLint.FunctionalTest.TestedProject` project, seems like it's dependant on some of the build output to run

jgardella:
@duckmatt Thanks for taking a look, I was having some trouble reproducing locally but was able to get it, and now I have the CI passing.

I will merge this PR, but I am curious if there is a reason behind why we use `dotnet-projinfo` to get the F# files from the project? Would it be sufficient to just pull the files from the proj file itself?

PR #394, opened 2020-03-23
Fix Docu (rule numbers, missing examples)

milbrandt:
- Fix rule numbers in docu (FL00xx instead of FS00xx)
- add missing configuration examples in FL0056 and FL0058 to stay in sync with other rules
- Increment heading only by one Level in FL0065

### DIFF
--- a/docs-gen/markdown/rules/FL0028.md
+++ b/docs-gen/markdown/rules/FL0028.md
@@ -14,3 +14,12 @@ Extract code out into private methods or functions.
 
 ## Rule Settings
 
+    [lang=javascript]
+    {
+        "MaxLinesInProperty": {
+            "enabled": false,
+            "config": {
+                "maxLines": 70
+            }
+        }
+    }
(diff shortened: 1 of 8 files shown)


docs-gen/markdown/rules/FL0028.md
```diff
@@ -14,3 +14,12 @@ Extract code out into private methods or functions.
 
 ## Rule Settings
 
+    [lang=javascript]
+    {
+        "MaxLinesInProperty": {
+            "enabled": false,
+            "config": {
+                "maxLines": 70
+            }
```
milbrandt: @duckmatt Json example extracted from `DefaultConfiguration.json`. Is this `config`required? If yes, then other docu pages Needs to be fixed (eg FL0029)

milbrandt: On testing with configuration of `maxCharactersOnLine` the `config` seems to be required.

duckmatt: Ideally it shouldn't be required and use the default if not specified, I'll take a look

duckmatt:
Awesome, thank you!

docs-gen/markdown/rules/FL0028.md
milbrandt: tested it with 0.13.3 - didn't test if the rework of #388 Change the behaviour

Issue #399, opened 2020-04-01
NetCore 3.1 support


akhansari:
### Description

FSharpLint doesn't support NetCore 3.1

### Repro steps

1. Only SDK 3.1 is installed (docker 3.1-alpine)
1. `dotnet tool restore` or `dotnet tool install dotnet-fsharplint`
1. `dotnet fsharplint lint <input>`
1. Error: The framework 'Microsoft.NETCore.App', version '2.1.0' was not found.

### Known workarounds

Also install the SDK 2.1

### Related information

* Operating system: alpine
* Version: 0.13.3

PR #400, opened 2020-04-01
Update paket to 5.243.0

milbrandt:
previous  5.194.3)

paket supports netcoreapp3.1 since 5.225.0 (thus will be required for #399)


milbrandt:
replaced by PR #401 which updates build Environment to .NET core 3.1

PR #414, opened 2020-04-21
Remove net461 workaround

milbrandt:
As net461 is no longer a target platform the workaround of Feb 2019 should no longer be required, should it?

Is the travis build still used or can the file removed completely?


duckmatt:
Yep, that should no longer be needed, thank you.

Regarding removing Travis, I think now with dotnet core that makes sense, not sure there's much value checking it compiles on both platforms now

Issue #419, opened 2020-05-09
Performance Regression


duckmatt:
### Description

Performance test is failing locally, previously ran under 200ms and now taking around 4s. It is a very large file the test operates on, so for your average file it's still operating somewhat efficiently. However no big default features have been added so there's no good reason for this degradation, we should look at getting it back where it was

### Repro steps

Run the performance tests locally

### Expected behavior

Performance test passes

### Actual behavior

Performance test fails

Issue #392, opened 2020-03-23
New rule proposal: Track uses of "TODO" tags


milbrandt:
### Description

TODO tags are commonly used to mark places where some more code is required, but which the developer wants to implement later.
Sometimes the developer will not have the time or will simply forget to get back to that tag.
This rule is meant to track those tags and to ensure that they do not go unnoticed.

### Expected behavior

Report `TODO`and synonymns, see at http://wordaligned.org/articles/todo

This rule is available in several static Code ananlyzers of other langes, e.g. in SonarQube (https://jira.sonarsource.com/browse/RSPEC-1135)

### Actual behavior

rule not available today

### Known workarounds

One might configure a hint

### Related information

Rule not listed at http://fsprojects.github.io/FSharpLint/Rules.html

duckmatt:
Thanks, good suggestion!

abelbraaksma:
You may find it interesting to check the code of F# Powertools. It's not supported after VS2017, but it supported task lists of TODO and FIXME and the like: https://github.com/fsprojects/Archive-VisualFSharpPowerTools/blob/master/README.md

xperiandri:
Now `TODO`s appear in the Task List tool window of VS 2022. I suppose we can close this?

milbrandt:
> Now `TODO`s appear in the Task List tool window of VS 2022. I suppose we can close this?

As I'm currently no longer using F# on my job, I can't test it. But if the TODO appear, it's a good improvement.

PR #422, opened 2020-05-13
Add BenchmarkDotnet project for profiling and fix performance issues

jgardella:
Given the recent performance regressions (#419), I have added a project using BenchmarkDotnet to use for profiling the linter. Currently, this is a very simple setup to just benchmark and profile the `lintParsedFile` function. It will produce an `etl` file which can be opened with `PerfView` to analyze the timing of the functions called. We may be able to set up automated performance tests using this in the future, but for now we can continue to use the existing one we have set up.

I fixed a couple of performance issues discovered using this:
- `toWarning` was being called for every lint suggestion and doing a string split on the entire file contents to get them by line. I pulled out this split and now do it only once.
- removed `addDefaults` in the naming rule helpers

With these fixes, the call to `lintParsedFile` seems to be around 500ms, which seems pretty reasonable given the file its operating on is pretty large. Based on the profiling, I think any further optimizations would have to be done in the lint rules themselves.


src/FSharpLint.Core/Application/Lint.fs
```diff
@@ -214,18 +216,18 @@ module Lint =
                 enabledRules.AstNodeRules |> Array.map (fun rule -> rule.Name)
             |] |> Array.concat |> Set.ofArray
 
-        let suppressionInfo = Suppression.parseSuppressionInfo allRuleNames lines
+        let suppressionInfo = Suppression.parseSuppressionInfo allRuleNames (Array.toList lines)
```
duckmatt, line 219: I'd be tempted to have `parseSuppressionInfo` operate on the array to avoid creating that list with it being quite large for big files

duckmatt:
Nice, 500ms does seem reasonable for that file. I converted `toWarning` locally the other night to seek to the line for similar performance, given the array of lines is also used elsewhere this probably makes some more sense. 

I removed the tokenisation to detect if the suppressions are within comments, this was adding around a second, I figured the chance of a suppression string appearing outside of a comment in a file outside of this project is pretty slim. What do you think? Alternatively we could have the clients pass in the tokens

jgardella:
Yes, I noticed that tokenization was causing a slow down as well. I considered having the client pass the tokens in, but given that there is not really a standard way to say "give me all the tokens for a F# file" (like how we can call one function to get an AST for the file), this seems like an unnecessary burden for clients.

I think your fix is good for now, but I am curious if there is a quicker way to get access to the tokens. For instance, I would think that our call to parse the AST of the file must be doing tokenization at some point; however this doesn't seem accessible through FCS.

duckmatt:
Yeah, that makes sense. Getting the AST must be parsing to the tokens at some point, I wonder if it's potentially doing it on a line by line basis while generating the tree though. Another potential alternative is to use the AST to see if the suppression is outside of the tree, although I'm not sure it'll be worth the effort - I'll be pretty surprised to see someone run into the issue

jgardella:
Yes, I agree it's probably not a big deal for this case. However it may be useful to have a more robust way of getting comments if we want to build lint rules based on them (e.g. #392).

I suspect you might be right about FCS doing tokenization line-by-line, but in that case they could add an optional argument for a function to call on each token, and we could then easily aggregate our own list of tokens to work with while still only tokenizing the file once.

duckmatt:
That's a good point, that todo rule would be quite a nice to have too. Although I'm thinking it's probably worth taking the approach of the client passing in the tokens. My concern regarding performance is for people using the linter as a library, mainly Ionide, they should I think already have the tokens available. For the console app CPU usage and performance isn't as key, we can always do the tokenisation in parallel to the AST - although having a function like you described in FCS would be very nice

jgardella:
Ok, I see your point. Basically, in our API we should provide a way for the client to pass in the tokens in case they already have them (the same as they can call `lintParsedFile` if they already have the AST). I think there is still value in having another function which does not take in any tokens and does the tokenization itself, for clients who don't have tokens already and don't want to have to do the tokenization themselves.

duckmatt:
Yeah, that makes sense. 

Off topic but thinking it might be worth us reviewing the defaults, `recursiveAsyncFunction` looks like it should be able to go by default, the formatting rules could be a bit more controversial

PR #431, opened 2020-05-23
Check record fields in TypedItemSpacing rule

jgardella:
Updates the TypedItemSpacing to check record fields as well. We should also check the types in union case fields, but there seems to be a bug in the parser where the range for the union case fields only includes the type, not the field name.


jgardella:
@duckmatt I made an update here to include Union Case fields in the syntax array. Could you verify that this makes sense? I assume the array should include any AstNodes we can find in the tree.

duckmatt:
Yep, that makes sense, I'm trying to look into removing this array (although starting with just removing this skip array for now - have a bank holiday weekend so hoping to knock it off shortly). I'm wondering for these formatting rules is there any way to leverage Fantomas?

Issue #435, opened 2020-05-27
Linter fails for multi-targeted projects


jgardella:
### Description

Linter fails for multi-targeted projects.

### Repro steps

Run the linter against a project with `<TargetFrameworks>` set. It will fail to lint with an MSBuild error.
This seems to be an issue with how we use `dotnet-proj-info`.

Krzysztof-Cieslak:
I'll investigate this tomorrow

PR #437, opened 2020-05-31
[WIP] Remove dotnet.projinfo

jgardella:
This PR removes the use of Dotnet.ProjInfo to load F# files from a project, instead loading them by parsing the project file.


src/FSharpLint.Core/Framework/ParseFile.fs
```diff
@@ -116,22 +89,43 @@ module ParseFile =
...
-        let source = File.ReadAllText(file)
+    let private parse file source (checker:FSharpChecker) =
+        let sourceText = SourceText.ofString source
+        let projectOptions = getProjectOptionsFromScript checker file source
+        let parsingOptions = checker.GetParsingOptionsFromProjectOptions projectOptions |> fst
+        let parseResults =
+            checker.ParseFile(file, sourceText, parsingOptions)
+            |> Async.RunSynchronously
 
-        let projectOptions =
```
jgardella, line 123: The main thing I'm not sure about in this PR is that we're now always calling `getProjectOptionsFromScript` instead of possibly using the options which resulted from the call to Dotnet.ProjInfo. Do we lose anything as a result?

Krzysztof-Cieslak:
I'm not a fan of this change. MsBuild files shouldn't be parsed as an XML - they can contain variable assignments or conditional statements and more. This means that by using some naive XML parsing you'll lose accuracy. Additionally `getProjectOptionsFromScript` doesn't work correctly on .Net Core - in Ionide we have lot of custom logic to resolve dependencies correctly for .Net Core in script context. @baronfel can tell more about that. 

If there are problems with `dotnet-proj-info` we should try fixing them rather than removing it here.

baronfel:
if you're doing anything with dll references for a project, `getProjectOptionsFromScript` will not give you correct references for any other target framework than `netstandard2.0`, because that's what FCS is compiled as, and the current code is decently pinned to that. In FSAutoComplete I had to write a .Net SDK-agnostic reference-finder to fill the gap here in FCS.  If you're not doing anything with the `-r` references for a project then you can ignore all of this :)

src/FSharpLint.Core/Framework/ParseFile.fs
baronfel: You should be worried here. There are things like compiler defines that come from the project file that won't be read correctly in this case. Target framework profiles, etc as well.

I have to ask, what's the main motivation for dropping DPI? In my opinion, it's the most consistent and correct view of the project file you're going to get, because it actually has msbuild evaluate the project file and collect the F# compiler arguments that the SDK would use to compile that file.

Krzysztof-Cieslak:
I do actually have other question - is `TypeCheckResults` used at all by FSharpLint? If all lints are based only on (untyped) AST then you don't need to type-check project at all, but only parse files, and you wouldn't need ProjectOptions at all, only ParseOptions. 

However it doesn't change my original point - I'd still use DPI to get ParseOptions - it's important to get the correct list of files and defines.

baronfel:
A couple concrete example of where parsing the project file wouldn't be enough to get the define list (not to belabor the point):

* If a project has a `Directory.Build.props` that adds defines, simple project file parsing won't show that
* If a project has globs for out-of-order but not-mutually dependent files, simple project file parsing won't show that easily without reimplementing globbing.

jgardella:
Thanks @Krzysztof-Cieslak and @baronfel for the feedback! This PR was basically an idea to remove a dependency and simplify how we get the files we need to run the linter on from a project. I had some suspicions this was a bad idea and was hoping someone would tell me why :). So thanks for that!

> I do actually have other question - is `TypeCheckResults` used at all by FSharpLint? If all lints are based only on (untyped) AST then you don't need to type-check project at all, but only parse files, and you wouldn't need ProjectOptions at all, only ParseOptions.

Yes, it is used for a small number of rules. For example, RedundantNewKeyword uses it to check if a type implements `IDisposable` to determine if `new` should be used when instantiating an instance.

The main issue which I was trying to solve with this is supporting multi-targeted projects (#435). I think because of the way we are using dotnet-projinfo these are not working. I also had a suspicion that calling MSBuild within the linter could cause issues when the linter is [run as an MSBuild task](http://fsprojects.github.io/FSharpLint/MSBuildTask.html), though I have no proof for that :)

src/FSharpLint.Core/Framework/ParseFile.fs
jgardella: See my other reply for more info, but basically just trying to support multi-targeted projects. I'm convinced doing things this way is a bad idea now, so if we can figure out what the issue is with how we're using DPI and multi-targeted projects, then I'm happy to keep using it

PR #433, opened 2020-05-27
[WIP] Rewrite build process

jgardella:

Issue #340, opened 2019-05-23
CLI tool: add flag to print ONLY warnings


Kelvin4702:
### Description

The current output of the CLI tool includes the lint progress and result for every file. It would be nice to have a way to filter out successful results and only print the warnings.

### Repro steps

1. dotnet fsharplint -f "foo.fsproj"

### Expected behavior

```
========== Linting File2.fs ==========
blah
========== Finished: 1 warnings ==========
========== Summary: 1 warnings ==========
```

### Actual behavior

```
========== Linting File1.fs ==========
========== Finished: 0 warnings ==========
========== Linting File2.fs ==========
blah
========== Finished: 1 warnings ==========
========== Linting File3.fs ==========
========== Finished: 0 warnings ==========
========== Summary: 1 warnings ==========
```

### Known workarounds

Piping the output into another tool, i.e. grep:
```
dotnet fsharplint -f "foo.fsproj" | grep -Pzo '(?s)=+ Linting (\N*) =+\n(?!=+ Finished: 0 warnings =+\n).*?=+ Finished: [^0] warnings =+\n'
```

### Related information 

* Operating system: Windows 10 Pro
* dotnet-fsharplint version: 0.11.1
* .NET Core version: 2.1.505

artempyanykh:
Curious, was there any change here?

Meantime kudos @Kelvin4702 for a workaround!

PR #438, opened 2020-06-03
[WIP] API redesign

jgardella:
Redesigning the public API and general code cleanup.

Still WIP.

**Summary**
- Separate `FSharpLine.Console` and `FSharpLint.Core` resources.
- Add `Resources.Format` function to remove boilerplate around using resources.
- Refactor `Lint.fs`.
   - All the publicly-exposed API functions now have a similar structure of building a list of `FileParseInfo`s based on the input, which is then passed to `lintFiles`.
   - `OptionalLintParameters` renamed to `LintParameters`.
   - Removed `LintInfo` type, replaced by  `LintParameters`.
   - Replaced `ReceivedWarning` and `ReportLinterProgress` in `OptionalLintParameters` with `HandleLintEvent` in `LintParameters`. This is a single callback function which clients can provide to learn when different `LintEvent`s occur in the linter.
- Deleted `Lint.fsi` signature file and added appropriate accessibility modifiers. Not 100% sure about this change, but one motivator was that the API docs produced in #433 by FSharp.Formatting were including things not defined in the signature file.
- Updated `FSharpLint.Console` project to work with API changes.
   - Now only produces output for files which have lint warnings. Previous behavior can be enabled with the `--all-files (-a)` flag. (#340).
- Expose diagnostic logging through the `handleLintEvent` callback. The LintEvent `LogMessage` indicates when the linter produces a log. This can be exposed in the dotnet tool with the `--verbose (-v)` flag.

**TODO**
- More generic support for passing parameters to MSBuild run on projects, aside from just `ReleaseConfiguration`.
- Timing logs for the `runAstRule` and `runLineRule` functions are a bit uninformative because they are called for each line/AST. Possibly we could aggregate the timing for each rule within a file before logging it.


duckmatt:
I'm going to close this for now as it has fallen behind, might potentially be worth us doing each refactor individually too to make it easier to get the changes in

PR #444, opened 2020-06-10
Update index.html

Krzysztof-Cieslak:
One of the fixes that were released after you've generated scaffold for docs.


duckmatt:
@jgardella updated!

PR #453, opened 2020-06-13
Add new rule NoPartialFunctions

jgardella:
Adds a new rule to disallow partial functions. The rule will suggest to use non-partial functions or pattern matching instead.


src/FSharpLint.Core/Rules/Conventions/NoPartialFunctions.fs
```diff
@@ -0,0 +1,108 @@
...
+type Config = {
+    AllowedPartials:string list
+    AdditionalPartials:string list
+}
+
+type private Replacement =
+    | PatternMatch
+    | Function of functionName:string
+
+let private partialFunctionIdentifiers =
```
jgardella, line 20: @Krzysztof-Cieslak I added all the partial collection functions that I'm aware of, anything else we should add?

knocte: @jgardella how about Map.Item -> Map.TryFind

knocte:
who needs to review this? would be nice to get it merged

duckmatt:
Just tried to resolve the conflicts in the Github UI, was not a good idea 😅

duckmatt:
@knocte Looks good to go in.

Potentially worth considering having these be hints although there is the downside which is that we'd lose the applicable documentation page

src/FSharpLint.Core/Rules/Conventions/NoPartialFunctions.fs
duckmatt: @knocte At the moment it'll require a little bit more work to handle methods on an object instance (some type checking) - might be worth another PR for that, we could potentially add:

`Map.find` -> `Map.tryFind`
`Map.findKey` -> `Map.tryFindKey`

Issue #464, opened 2020-10-16
Error with Nested Statements (FL0015) and multiple methods of records


davidtgillard:
### Description

The number of nested statements appears to not be  decremented after exiting the scope of a method within the body of a record.

E.g. running the linter on the following snippet

```
  type CoercedInt =
    { Value: int }
    member this.Create1 i = { Value=i }
    member this.Create2 (i: float) = { Value=Convert.ToInt32(i)}
    member this.Create3 (i: float32) = { Value=Convert.ToInt32(i)}
    member this.Create4 (i: bool) = { Value=Convert.ToInt32(i)}
    member this.Create5 (i: string) = { Value=Convert.ToInt32(i)}
    member this.Create6 (i: uint32) = { Value=Convert.ToInt32(i)}
    member this.Create7 (i: uint16) = { Value=Convert.ToInt32(i)}
    member this.Create8 (i: byte) = { Value=Convert.ToInt32(i)}
    member this.Create9 (i: char) = { Value=Convert.ToInt32(i)}
```

with the configuration 

```
{
"nestedStatements": {
        "enabled": true,
        "config": {
            "depth": 8
        }
    }
}
```
yields the following output:

> ========== Linting Program.fs ==========
> Code suggested not to be nested more deeply than a depth of 8.
> Error on line 60 starting at column 11
>     member this.Create9 (i: char) = { Value=Convert.ToInt32(i)}
>            ^
> --------------------------------------------------------------------------------
> Code suggested not to be nested more deeply than a depth of 8.

The issue also occurs with static methods and struct records.

### Repro steps

Run the linter (version 0.16.5) with the input as described above.

### Expected behavior

No deep nesting errors reported.

### Actual behavior

As above.

### Known workarounds

None.

### Related information

* Operating system: Windows 10, Ubuntu 20.04
* FSharpLinter version 0.16.5, installed as global DotNet tool

PR #470, opened 2020-11-22
ensure record members do not count as nested statements (#464)

davidtgillard:
Fix for #464. Tests included.


duckmatt:
Great, thank you!

PR #471, opened 2021-01-02
README&docs: fix broken links

knocte:


knocte:
@duckmatt FYI there's also this broken link in the README: http://fsprojects.github.io/FSharpLint/Rules.html (but I don't know how to fix this one).

duckmatt:
Great, thank you 👍. Thanks for the spot RE the other broken link in the readme, I think that can just be removed I'll push that shortly

knocte:
Actually my docs change didn't fix it :( and I can now see that all internal links in that page are broken, not the one I just changed.

PR #475, opened 2021-01-14
Improve "Remove broken link from readme"

knocte:
This improves commit 62e181d7fde15ff6cbd4a52c8bf4fb29e9ac0a96 by
using this technique:

https://stackoverflow.com/a/7335259/1623521

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ FSharpLint can be used in several ways:
 
 ## Documentation
 
-The [docs](http://fsprojects.github.io/FSharpLint/) contains an overview of the tool and how to use it.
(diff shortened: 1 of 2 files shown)


duckmatt:
Ah nice, thank you

knocte:
Actuallly it doesn't work 😭 so it needs to be reverted

knocte:
Unless you need to do something to update the docs website? not sure

duckmatt:
Ah sorry, I think I forgot to reply on the previous PR, the docs do currently get pushed on release (via tag), I'll have a look at pushing that shortly

knocte:
Ah great sounds good.

PR #476, opened 2021-01-15
Fix signature warnings for argument names

deviousasti:


duckmatt:
Thank you, are these warnings an IDE's spellcheck?

deviousasti:
They're mismatches between the argument names in the signature files and the implementation file's name, which throws warnings.
Thanks.

duckmatt:
Ah I see, thank you for the fix 👍

PR #477, opened 2021-02-04
Fix off by one error

LiteracyFanatic:
Exception occurs in line 90 of Rules.fs when TrailingNewLineInFile (FL0063) is set to true. Adjust the position of the error to fix the
issue.

Conceptually we tend to think of `\n` as starting a newline because most editors will display an empty line at the end in the presence of a trailing `\n`. In reality, `\n` simply separates lines and the next line does not exist until we add text to it. You can actually observe this behavior in vim which doesn't display an empty line but can be configured to show the trailing `\n`.


duckmatt:
Cool, that makes sense, thank you for the contribution!

PR #478, opened 2021-02-11
Package and API updates for FCS 39

baronfel:
This PR updates the packages for FCS and ProjectInfo to FCS-39-compatible versions.

The primary changes that impacted this project are:

* namespace changes for Range/Pos/PrettyNaming
* some members are no longer Async
* Dotnet.ProjectInfo is no longer maintained (it's now Ionide.ProjectInfo) so that switch was made as well.

As a result of the change to Dotnet.ProjectInfo, the projects had to be changed to target net5.0 instead of netstandard2.0. This is primary due to the use of the MSBuild Locators.

As a result of the removal of Async from certain members, the Async<bool> option members for some checks instead became `(unit -> bool) option`, to represent that same 'delayed' computation.


deviousasti:
I didn't see this PR, and went about it in parallel. d'oh!

image

Our code mostly looks the same, except that I opted to use `Lazy<bool>` instead of `unit -> bool`.

baronfel:
the test failures need to be addressed. I think the current test strategy for testing the dotnet tool will no longer work (due to packaging requirements for the msbuild locator) so the tests need to be updated to actually install/spawn the dotnet tool, similar to what the analyzer sdk, miniscaffold test, etc all have to do.

duckmatt:
Sorry for the slow response, I'll try to take a look at updating the tests to get it building over the next couple of evenings

baronfel:
:wave: is there anything I can help with here?

faldor20:
Hey, @duckmatt not to be annoying, but for all of us using ionide, this PR being blocked stops us from getting linting.
If you could see if these tests could be fixed that would be a huge help :)
Thanks for all your great work.

duckmatt:
Sorry all, I'll get on this. Have been a bit distracted from the project for a while for various reasons, looking to pick things up a bit more again

@baronfel do you know why the msbuild locator requires the dotnet tool to be spawned? The functional tests would be better run this way so I'll definitely look at sorting those, but for the console app tests it's quite handy directly calling the main function to easily debug. I've managed to get the console app tests passing: https://github.com/fsprojects/FSharpLint/pull/482/commits/771bd2a51b7627e14a3e92c6f5d0393428c82c79 but this is the first time I've really looked into the locator and I'm not seeing much docs around it so not completely sure if this is the right thing to be doing here

duckmatt:
Merged in via #482 thank you for the PR! Sorry it was such a slow merge

baronfel:
oh awesome @duckmatt! Thanks for finishing this up.  The locator seems to be set up that way because it defaults to locating .net sdk installs by running `dotnet --info`, which is a) hilarious and b) seems ripe for short-circuiting with an env variable like DOTNET_ROOT or something.

Issue #493, opened 2021-08-16
Should split nonPublicValuesNames into privateValuesNames and internalValuesNames


knocte:
I found this in the default fsharplint.json:

```
    "nonPublicValuesNames": {
        "enabled": true,
        "config": {
            "naming": "CamelCase",
            "underscores": "AllowPrefix"
        }
    },
```

With the above setting, an FSharpLint user cannot choose CamelCase for private values and PascalCase for internal values.

PR #495, opened 2021-09-02
Separate non public values rule in two rules

knocte:
Fixes https://github.com/fsprojects/FSharpLint/issues/493


duckmatt:
Awesome, looks great, thank you! Will try to get a new version pushed out shortly

PR #499, opened 2021-09-14
docs: try out <s> inside <a> instead of outside

knocte:
The previous attempt in PR #497 didn't work.


duckmatt:
Thank you - worth another go, might be that Fornax doesn't yet support strikethrough from markdown

PR #501, opened 2021-09-17
NoPartialFunctions: stop suggesting tryLast instead of tail

knocte:
tryLast doesn't have the same goal as tail; let's rather suggest
FSharpx.Collections.Seq.tryHeadTail extension for now, which has
just been merged [1] and will be in 3.0.0 release (at least until
we have a tryTail of some sort [2]).

[1] https://github.com/fsprojects/FSharpx.Collections/pull/176
[2] https://github.com/fsharp/fslang-suggestions/issues/803


duckmatt:
Thank you and good point 👍 . Interestingly I think this is the first time we're pointing to a third party lib in a suggestion, makes sense though

PR #506, opened 2021-09-21
Add new rule FavourTypedIgnore

su8898:
See rationale: https://twitter.com/ChetHusk/status/1351156012590526474


duckmatt:
Looks good and is a good addition! My only concern is making whether it should be a default, it's not something I've seen that widespread and will bring a lot of warnings to a lot of projects when it makes its way down to ionide. How do you feel about it being something users can decide to turn on?

su8898:
@duckmatt Many thanks for the suggestion. I've changed the default setting to false for favourTypedIgnore and updated the PR. Could you please have a look?

duckmatt:
@su8898 Great thank you! sorry for the slow response on this, I've ended up losing access temporarily due to not having 2fa on my account previously 😅, I'll get this merged in and released as soon as I get access back

Issue #195, opened 2016-12-11
cyclomatic complexity removed?


jmecosta:
hi @duckmatt 

notice that the http://fsprojects.github.io/FSharpLint/CyclomaticComplexity.html has been removed from code. Is there a new replacement for it?

regards

jorge costa

duckmatt:
Hi Jorge, 

It was removed during a reworking of the core of the linter, didn't seem like it was being used by any one so didn't deem it worthy of porting it across. Did you find it useful to have?

jmecosta:
i was updating the F# sonar plugin and being complexity such a core metric in sonar its curious that no one is using it in F#. usually complexity gives good indication where we start refactoring and putting limits on it its good to keep the code simple to follow. 

maybe it does not apply to F#?

duckmatt:
it can apply to f# but is certainly less useful than when applied to imperative languages, a pattern match against each case of a DU of *n* cases increases the CC of the func by *n*, with this it's easy to get a lot of "false positives" where refactoring the code to reduce the CC would probably make it more complex and less idomatic

jmecosta:
OK @duckmatt, thanks for the explanation. I will follow this recommendation also

PR #507, opened 2021-09-28
Added cyclomatic complexity rule

davidtgillard:
I know that a decision was made in the past to remove cyclomatic complexity from FSharpLint (see #195). However, cyclomatic complexity can be a useful quality metric and is required as part of many software assurance programs in industry. It would be extremely useful to myself and my co-workers if it were part of FSharpLint.

The original implementation included an option to not include match cases. I have omitted this option, since while discriminated unions and exhaustive pattern matching are much safer than, say, switch statements, they still increase cyclomatic complexity (even if they also happen to do an excellent job of managing this complexity).

### DIFF
--- a/docs/content/how-tos/rule-configuration.md
+++ b/docs/content/how-tos/rule-configuration.md
@@ -110,4 +110,5 @@ The following rules can be specified for linting.
 - [PrivateValuesNames (FL0067)](rules/FL0067.html)
 - [InternalValuesNames (FL0068)](rules/FL0068.html)
 - [GenericTypesNames (FL0069)](rules/FL0069.html)
+- [FavourTypedIgnore (FL0070)](rules/FL0070.html)
 - [CyclomaticComplexity (FL0071)](rules/FL0071.html)
--- a/docs/content/how-tos/rules/FL0070.md
+++ b/docs/content/how-tos/rules/FL0070.md
@@ -1,0 +1,30 @@
+---
+title: FL0070
+category: how-to
+hide_menu: true
+---
+
+# FavourTypedIgnore (FL0070)
+
+*Introduced in `0.20.2`*
+
+## Cause
+
+Use of `ignore` function is untyped.
+
+## Rationale
+
+Specifying a type prevents mistakes and aides maintainability.
+
+## How To Fix
+
+Add the type of the element being ignored as a type parameter to the `ignore`
+invokation, e.g. `ignore<string>`.
+
+## Rule Settings
+
+    {
+        "favourTypedIgnore": {
+           "enabled": true
+        }
+    }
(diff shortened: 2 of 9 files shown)


tests/FSharpLint.Core.Tests/Rules/Conventions/CyclomaticComplexity.fs
```diff
@@ -0,0 +1,258 @@
...
+    /// Verifies that an error is raised on a single match clause with multiple logical operators.
+    [<Test>]
+    member this.TestMultipleLogicalOperatorsInSingleMatchClause() =
+        let indent = TestConventionsCyclomaticComplexity.Indent
+        let newline = TestConventionsCyclomaticComplexity.NewLine
+        let code = "Module Program" + newline + 
+                   "let f() = " + newline +
+                   indent + "match \"dummyString\" with" + newline +
+                   indent + "| x when x = \"1\"" + ([2..maxComplexity+1] |> List.map (sprintf "|| x = \"%d\"") |> String.concat " ") + " -> ()" + newline +
+                   indent + "| _ -> ()"
```
knocte, line 183: question, why do you use things like `newline` and `indent` here instead of having a multi-line string where you place them yourself? (the latter would be much more readable, and you can still interpolate other things with %s AFAIU)

davidtgillard: Yes, okay, agreed. I will make that change.

Are there any other changes you'd like?

knocte: > Yes, okay, agreed. I will make that change.

Cool.

> Are there any other changes you'd like?

Not from my side. I'm not the maintainer, just started contributing recently.

knocte:
CI passing on Linux but failing on Windows? interesting...

davidtgillard:
I’ll try to sort this out on a windows machine.

davidtgillard:
Aha. So, it's a platform newline thing. Triple-quoted strings on Windows substitute '\n' for newlines, instead of Environment.NewLine = '\r\n'.

Check it out: 
```fsharp
let tripleQuotedString = """1
2
3"""
let concatenatedString = "1" + Environment.NewLine + "2" + Environment.NewLine + "3";;
printfn "%A" (Seq.toList tripleQuotedString) // yields ['1'; '\010'; '2'; '\010'; '3']
printfn "%A" (Seq.toList concatenatedString ) // yields ['1'; '\013'; '\010'; '2'; '\013'; '\010'; '3']
```

This is...surprising to me, and doesn't seem to be explicitly described in the F# language spec. Is it suprising to you?

At any rate, at least I know how to fix these tests for Windows.

duckmatt:
Awesome thanks, looks good!

Regarding the new lines what I think might be happening is that it's taking whatever characters are there in the file, with this file likely being LF (haven't double checked but it would seem reasonable if this is it)

davidtgillard:
My oversight - this behavior of triple-quoted strings is [well-defined](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/strings#triple-quoted-strings).

Thanks both for the feedback.

duckmatt:
Awesome, thank you 👍

Issue #509, opened 2021-10-04
Feature: rule to suggest use of "reraise()" instead of "raise"


knocte:
### Description

Using `raise ex` to re-raise an exception is bad in the same way as using `throw ex;` is bad in C# (instead of just `throw;`).

### Repro steps

Good example:

```
try
    foo ()
with
| ex ->
    if someCondition then
        reraise()
```

Bad example:

```
try
    foo ()
with
| ex ->
    if someCondition then
        raise ex
```

PR #516, opened 2021-10-19
Add new rule FavourReRaise

su8898:
Fixes https://github.com/fsprojects/FSharpLint/issues/509

### DIFF
--- a/docs/content/how-tos/rules/FL0073.md
+++ b/docs/content/how-tos/rules/FL0073.md
@@ -1,0 +1,29 @@
+---
+title: FL0073
+category: how-to
+hide_menu: true
+---
+
+# FailwithBadUsage (FL0073)
+
+*Introduced in `0.20.3`*
+
+## Cause
+
+Rule to detect usage of `raise ex` instead of `reraise()` inside `try...with|ex` blocks.
+
+## Rationale
+
+Using `raise ex` to re-raise an exception is bad as it loses valuable stacktrace information from the original exception.
+
+## How To Fix
+
+Use `reraise()` instead of `raise ex` inside `try...with|ex` blocks.
+
+## Rule Settings
+
+    {
+        "favourReRaise": {
+           "enabled": true
+        }
+    }
--- a/src/FSharpLint.Core/Rules/Conventions/FavourReRaise.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/FavourReRaise.fs
@@ -1,0 +1,51 @@
+module FSharpLint.Rules.FavourReRaise
+
+open System
+open FSharpLint.Framework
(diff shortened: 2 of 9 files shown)


docs/content/how-tos/rules/FL0073.md
```diff
@@ -0,0 +1,29 @@
...
+hide_menu: true
+---
+
+# FailwithBadUsage (FL0073)
+
+*Introduced in `0.20.3`*
+
+## Cause
+
+Rule to detect usage of `raise ex` instead of `reraise()` inside `try...with` blocks.
```
knocte, line 13: let's replace `try...with` with `try...with|ex`, as you mentioned `ex` before.

docs/content/how-tos/rules/FL0073.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Rationale
+
+Using `raise ex` to re-raise an exception is bad as it loses valuable stacktrace information from the original exception.
+
+## How To Fix
+
+Use `reraise()` instead of `raise ex` inside `try...with` blocks.
```
knocte, line 21: same here

src/FSharpLint.Core/Rules/Conventions/FavourReRaise.fs
```diff
@@ -0,0 +1,51 @@
...
+open FSharp.Compiler.Syntax
+open FSharpLint.Framework.Ast
+open FSharpLint.Framework.Rules
+
+let private runner (args: AstNodeRuleParams) =
+    let generateError range =
+        { Range = range
+          Message = Resources.GetString "RulesFavourReRaise"
+          SuggestedFix = None
+          TypeChecks = [] }
```
knocte, line 15: use List.Empty here please

src/FSharpLint.Core/Text.resx
```diff
@@ -327,4 +327,7 @@
   <data name="RulesFailwithBadUsage" xml:space="preserve">
     <value>Bad usage of failwith: {0}.</value>
   </data>
+  <data name="RulesFavourReRaise" xml:space="preserve">
+    <value>Use reraise() instead of raise.</value>
```
knocte, line 331: this should be more specific because this rule doesn't advocate replacing all uses of `raise` with `reraise`, only the ones that use the same captured exception

su8898: How about `Use reraise() instead of using raise with the same captured exception.`?

knocte: Sure but add the word "Rather" before the word "use"

su8898:
@knocte Hi, I've updated the PR with your suggestions. Could you please have a look? thanks

Issue #519, opened 2021-11-10
Support for dotnet 6.0


ethangk:
### Description

Given the [recent](https://dotnet.microsoft.com/download/dotnet/6.0) release of dotnet 6.0, I've attempted to bump my project to that version. We're running into issues with FSharpLint not supporting dotnet 6.0, from what we can see.

### Repro steps

Bump the dotnet version in the docker container to dotnet 6.0. `dotnet tool restore` works ok, but running:
`dotnet fsharplint lint projectName.fsproj` results in

```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '5.0.0' (x64) was not found.
  - The following frameworks were found:
      6.0.0 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=5.0.0&arch=x64&rid=debian.11-x64
```

This is using the most recent (`0.20.2`) version of FSharpLint.

Thanks!

ethangk:
Thanks for getting back to me so quickly. I've cloned the project locally and added the flag to the `global.json` then built. I'm trying to reference it from my project (by adding the local path as an additional source), so that I can test and it's still not working. I suspect it's using the nuget version instead of this local version. I'm fairly new to the dotnet package world. What's the easiest way to verify this/ force it to use the local version?

knocte:
I think you need to generate a nupkg file and use it locally, but don't remember from the top of my head how to do it (especially taking in account that Paket is being used).

Cheroukee:
My team has the same issue as @ethangk 
I've taken to trying the proposed fix, by building it locally and copying the output folder too a container that runs only net6, without any success. 

After doing some research I fell on this thread :  https://github.com/dotnet/sdk/issues/16427 

Setting this excerpt in the project file for the tool made it run on a 6.x runtime
```
<PropertyGroup>
  <RollForward>Major</RollForward>
</PropertyGroup>
```

From what I observed, doing the change in the `config.json` file only changed the behaviour at build time, where if I ; `rollForward: latestMajor` it will build using 6.x for 5.x, but when trying to launch the tool, it fails with the original error.

I can create a PR with the unique change if you think this is a valid fix. Otherwise we can also upgrade to 6.x / multitarget. What's your view on this ?

[...]

PR #520, opened 2021-11-15
Enable running tool on .Net 6

Cheroukee:
This will fix issue #519 

This adds a rollForward "Major" to the console application project file, enabling the app to run on Net,6 

This was tested by running the tool from inside a container based on the MSFT for net. 6, after copying the output from the project into the image.  Without the fix, this method would fail.

### DIFF
--- a/src/FSharpLint.Console/FSharpLint.Console.fsproj
+++ b/src/FSharpLint.Console/FSharpLint.Console.fsproj
@@ -12,6 +12,7 @@
     <AssemblyName>dotnet-fsharplint</AssemblyName>
     <RootNamespace>FSharpLint.Console</RootNamespace>
     <IsPackable>true</IsPackable>
+    <RollForward>Major</RollForward>
   </PropertyGroup>
 
   <ItemGroup>


ethangk:
Nice one, thanks for looking into this

knocte:
Thanks @Cheroukee! Before I merge tho, I'm going to bother you with 2 nitpicks, if you don't mind, in order not to pollute git-blame:
- Can you remove the extra unrelated commit (the 2nd one)?
- Can you add the text "Fixes #519" in the commit message body of your first commit?
Thanks again.

Cheroukee:
I goofed on the first rebase but now it's all fine

knocte:
Thanks! Will release a v0.20.3 soonish.

knocte:
0.21.0 was released today, including this change; please test it and report any issues you find.

Also if any of you want to contribute more, I see how we can improve this solution: instead of rolling forward to .NET6.0, we should actually switch our infrastructure to .NET6 (because it's LTS, unlike 5.0), and roll backward instead of roll forward (to maintain compatibility with 5.0).

Cheroukee:
I can take a look at that. I feel like it would be a good idea to do that indeed

PR #522, opened 2021-11-22
Add 2 new rules NestedFunctionNames and UnnestedFunctionNames

su8898:
These rules allow configuring naming conventions for nested and unnested function names.

### DIFF
--- a/docs/content/how-tos/rules/FL0075.md
+++ b/docs/content/how-tos/rules/FL0075.md
@@ -1,0 +1,33 @@
+---
+title: FL0075
+category: how-to
+hide_menu: true
+---
+
+# UnnestedFunctionNames (FL0075)
+
+*Introduced in `0.21.1`*
+
+## Cause
+
+Unnested function naming does not match the specified config.
+
+## Rationale
+
+Consistency aides readability.
+
+## How To Fix
+
+Update the unnested function names to be consistent with the rules you have specified
+
+## Rule Settings
+
+    {
+        "unnestedFunctionNames": {
+            "enabled": false,
+            "config": {
+                "naming": "PascalCase",
+                "underscores": "None"
+            }
+        }
+    }
--- a/docs/content/how-tos/rules/FL0076.md
+++ b/docs/content/how-tos/rules/FL0076.md
@@ -1,0 +1,33 @@
+---
+title: FL0076
+category: how-to
+hide_menu: true
+---
+
+# NestedFunctionNames (FL0076)
+
+*Introduced in `0.21.1`*
+
+## Cause
+
+Nested function naming does not match the specified config.
+
+## Rationale
+
+Consistency aides readability.
+
+## How To Fix
+
+Update the Nested function names to be consistent with the rules you have specified
+
+## Rule Settings
+
+    {
+        "NestedFunctionNames": {
+        "enabled": false,
+            "config": {
+                "naming": "CamelCase",
+                "underscores": "None"
+            }
+        }
+    }
--- a/src/FSharpLint.Core/Rules/Conventions/Naming/NamingHelper.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/Naming/NamingHelper.fs
@@ -361,3 +361,22 @@ let rec identFromSimplePat = function
     | SynSimplePat.Id(ident, _, _, _, _, _) -> Some(ident)
     | SynSimplePat.Typed(p, _, _) -> identFromSimplePat p
     | SynSimplePat.Attrib(_) -> None
+
+let rec isNested args nodeIndex =
+    let parent = args.SyntaxArray.[nodeIndex].ParentIndex
+    let actual = args.SyntaxArray.[parent].Actual
+
+    match actual with
+    | AstNode.Expression (SynExpr.LetOrUse (_, _, _, _, _)) -> true
+    | _ -> false
+
+let getFunctionIdents _ =
+    function
+    | SynPat.LongIdent (longIdent, _, _, pats, _, _) ->
+        match pats with
+        | SynArgPats.Pats _ ->
+            match List.tryLast longIdent.Lid with
+            | Some ident -> (ident, ident.idText, None) |> Array.singleton
+            | None -> Array.empty
+        | _ -> Array.empty
+    | _ -> Array.empty
--- a/src/FSharpLint.Core/Rules/Conventions/Naming/NestedFunctionNames.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/Naming/NestedFunctionNames.fs
@@ -1,0 +1,25 @@
+module FSharpLint.Rules.NestedFunctionNames
+
+open FSharp.Compiler.Syntax
+open FSharpLint.Framework.Ast
+open FSharpLint.Framework.AstInfo
+open FSharpLint.Framework.Rules
+open FSharpLint.Rules.Helper.Naming
+
+let private getIdentifiers (args: AstNodeRuleParams) =
+    match args.AstNode with
+    | AstNode.Binding (SynBinding (_, _, _, _, _attributes, _, _, pattern, _, _, _, _)) ->
+        if isNested args args.NodeIndex then
+            getPatternIdents Accessibility.Public getFunctionIdents true pattern
+        else
+            Array.empty
+    | _ -> Array.empty
+
+let rule config =
+    { Name = "NestedFunctionNames"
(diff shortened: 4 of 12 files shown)


docs/content/how-tos/rules/FL0075.md
```diff
@@ -0,0 +1,32 @@
+---
+title: FL0075
+category: how-to
+hide_menu: true
+---
+
+# UnnestedFunctionNames (FL0075)
+
+*Introduced in `0.20.3`*
```
siwatanejo, line 9: please change this to 0.21.1

docs/content/how-tos/rules/FL0076.md
```diff
@@ -0,0 +1,32 @@
+---
+title: FL0076
+category: how-to
+hide_menu: true
+---
+
+# NestedFunctionNames (FL0076)
+
+*Introduced in `0.20.3`*
```
siwatanejo, line 9: same here

docs/content/how-tos/rules/FL0075.md
```diff
@@ -0,0 +1,32 @@
...
+## Rule Settings
+
+    {
+        "unnestedFunctionNames": {
+        "enabled": false,
+        "config": {
+            "naming": "PascalCase",
+            "underscores": "None"
+        }
+    }
```
knocte, line 32: please remove trailing whitespace from the line above

docs/content/how-tos/rules/FL0076.md
```diff
@@ -0,0 +1,32 @@
...
+## Rule Settings
+
+    {
+        "NestedFunctionNames": {
+        "enabled": false,
+        "config": {
+            "naming": "CamelCase",
+            "underscores": "None"
+        }
+    }
```
knocte, line 32: same here

src/FSharpLint.Core/Rules/Conventions/Naming/NamingHelper.fs
```diff
@@ -361,3 +361,14 @@ let rec identFromSimplePat = function
...
+
+let rec isNested args nodeIndex =
+    let parent = args.SyntaxArray.[nodeIndex].ParentIndex
+    let actual = args.SyntaxArray.[parent].Actual
+
+    match actual with
+    | AstNode.ModuleDeclaration (SynModuleDecl.Types (_, _)) -> false
+    | AstNode.ModuleDeclaration (SynModuleDecl.NestedModule (_, _, _, _, _)) -> false
+    | AstNode.Binding (SynBinding (_, _, _, _, _, _, _, _, _, _, _, _)) -> false
+    | AstNode.Expression (SynExpr.LetOrUse (_, _, _, _, _)) -> true
```
knocte, line 373: this match block can be hugely simplified if you move the case `AstNode.Expression (SynExpr.LetOrUse (_, _, _, _, _))` to the first position, and then just use `_ -> false` as a second case

src/FSharpLint.Core/Rules/Conventions/Naming/UnnestedFunctionNames.fs
```diff
@@ -0,0 +1,35 @@
...
+            match List.tryLast longIdent.Lid with
+            | Some ident -> (ident, ident.idText, None) |> Array.singleton
+            | None -> Array.empty
+        | _ -> Array.empty
+    | _ -> Array.empty
+
+let private getIdentifiers (args: AstNodeRuleParams) =
+    match args.AstNode with
+    | AstNode.Binding (SynBinding (_, _, _, _, _attributes, _, _, pattern, _, _, _, _)) ->
+        match isNested args args.NodeIndex with
```
knocte, line 23: please don't use match blocks for true/false values, just use an `if ... else`

su8898:
@knocte thanks for the review. I've updated the PR with your suggestions. Could you please take a look?

src/FSharpLint.Core/Rules/Conventions/Naming/NestedFunctionNames.fs
```diff
@@ -0,0 +1,35 @@
...
+            match List.tryLast longIdent.Lid with
+            | Some ident -> (ident, ident.idText, None) |> Array.singleton
+            | None -> Array.empty
+        | _ -> Array.empty
+    | _ -> Array.empty
+
+let private getIdentifiers (args: AstNodeRuleParams) =
+    match args.AstNode with
+    | AstNode.Binding (SynBinding (_, _, _, _, _attributes, _, _, pattern, _, _, _, _)) ->
+        match isNested args args.NodeIndex with
```
knocte, line 23: this is another match block on true/false, please convert into if else

su8898: I've updated the PR with this fix. Could you please take a look? thanks

src/FSharpLint.Core/Rules/Conventions/Naming/UnnestedFunctionNames.fs
```diff
@@ -0,0 +1,36 @@
...
+let private getFunctionIdents _ =
+    function
+    | SynPat.LongIdent (longIdent, _, _, pats, _, _) ->
+        match pats with
+        | SynArgPats.Pats _ ->
+            match List.tryLast longIdent.Lid with
+            | Some ident -> (ident, ident.idText, None) |> Array.singleton
+            | None -> Array.empty
+        | _ -> Array.empty
+    | _ -> Array.empty
```
knocte, line 18: this function is exactly the same in UnnestedFunctionNames.fs and NestedFunctionNames.fs, let's place it in a common place to share it between the two

su8898: I've placed the function in a common place and updated the PR. Could you please take a look? thanks

src/FSharpLint.Core/Rules/Conventions/Naming/NamingHelper.fs
```diff
@@ -361,3 +361,22 @@ let rec identFromSimplePat = function
...
+let rec isNested args nodeIndex =
+    let parent = args.SyntaxArray.[nodeIndex].ParentIndex
+    let actual = args.SyntaxArray.[parent].Actual
+
+    match actual with
+    | AstNode.Expression (SynExpr.LetOrUse (_, _, _, _, _)) -> true
+    | _ -> false
+
+let getFunctionIdents _ =
+    function
```
knocte, line 374: I'm not a fan of using the keyword `function` here, can you just write a normal function?

src/FSharpLint.Core/Rules/Conventions/Naming/UnnestedFunctionNames.fs
```diff
@@ -0,0 +1,25 @@
...
+open FSharpLint.Framework.Rules
+open FSharpLint.Rules.Helper.Naming
+
+let private getIdentifiers (args: AstNodeRuleParams) =
+    match args.AstNode with
+    | AstNode.Binding (SynBinding (_, _, _, _, _attributes, _, _, pattern, _, _, _, _)) ->
+        if isNested args args.NodeIndex then
+            Array.empty
+        else
+            getPatternIdents Accessibility.Public getFunctionIdents true pattern
```
knocte, line 15: are we sure we want to use Accessibility.Public here? we want all functions, even private and internal ones

su8898: I don't think the `Accessibility.Public` is used in this case. You think it's best to add a `Both` or `NotApplicable` member to the `Accessibility` DU?

knocte: I don't understand how can this be not used?

webwarrior-ws: Passing `Accessibility.Public` here means we start within public scope.
So it's `getFunctionIdents` function that can decide what is included based on identifier's accessibility.
Which ignores accessibility parameter, and processes all functions.

src/FSharpLint.Core/Rules/Conventions/Naming/NamingHelper.fs
webwarrior-ws: Yes, and better to rename `_` to `accessibility` or `_accessibility` just to see what is being ignored

Issue #525, opened 2021-12-15
Should favour names in patterns and DU subtypes


knocte:
There are two ways to write things, named or unnamed.

Unnamed:

```
type Data =
    | TwoParts of string * string
    | OnePart of string

let examineData x =
    match data with
    | OnePart p1 -> p1
    | TwoPart (p1, p2) -> p1 + p2
```

Named:

```
type Data =
    | TwoParts of part1: string * part2: string
    | OnePart of part1: string

let examineData x =
    match data with
    | OnePartData(part1=p1) -> p1
    | TwoPartData(part1=p1; part2=p2) -> p1 + p2
```

We could have a rule that warns about the former and prefers the latter.

PR #528, opened 2021-12-16
Add new rule FavourNamedMembers

janus:
Fixes https://github.com/fsprojects/FSharpLint/issues/525


knocte:
Rebased in https://github.com/fsprojects/FSharpLint/pull/583

Issue #526, opened 2021-12-15
Should favour use of singleton function


knocte:
Instead of

```
let foo = [ x ]
let bar = [| x |]
```

There should be a rule that prefers this:

```
let foo = List.singleton x
let bar = Array.singleton x
```

PR #534, opened 2021-12-20
Add new rule FavourSingleton

su8898:
Fixes https://github.com/fsprojects/FSharpLint/issues/526

### DIFF
--- a/docs/content/how-tos/rule-configuration.md
+++ b/docs/content/how-tos/rule-configuration.md
@@ -115,3 +115,4 @@ The following rules can be specified for linting.
 - [FailwithBadUsage (FL0072)](rules/FL0072.html)
 - [FavourReRaise (FL0073)](rules/FL0073.html)
 - [FavourConsistentThis (FL0074)](rules/FL0074.html)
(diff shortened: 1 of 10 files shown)


duckmatt:
I think it may be possible to express this as one of the hints, although I'm not averse to an additional rule either. Trade off is the rule is easier to ignore one rule specifically, but adding an additional hint should have better performance over an additional rule

(Also apologies, just merged in another PR so needs conflicts resolved)

knocte:
Disadvantages of converting this into a hint:
1. Can't be disabled by default (and I've already known 2 people which don't like this rule).
2. There's already a hint that recommends you to use [x] instead of x::[], so fixing this hint might lead to yet another hint, unless we change that hing to recommend a Singleton func, but should we? (See 1. :) )

knocte:
> ...just merged in another PR so needs conflicts resolved)

No worries, I'll rebase myself soon.

knocte:
Superseded by https://github.com/fsprojects/FSharpLint/pull/587

PR #538, opened 2022-03-10
Add new rule AvoidTypeHintSuffixesInNames

janus:
Rule to detect type member name suffixes with type hints.


knocte:
Superseded by https://github.com/fsprojects/FSharpLint/pull/585

Issue #539, opened 2022-03-18
`failwith` rules produce incorrect warning when throwing in `Option.defaultWith`


peterhirn:
### Description

`failwith` rules produce incorrect warning when throwing in `Option.defaultWith`.

### Repro steps

```fsharp
let failIfNone (opt: 'a option) : 'a =
    opt
    |> Option.defaultWith (fun _ -> failwith "A unique error message")
```

### Actual behavior

```
Bad usage of failwith: Consider using unique error messages with failwith.
Error on line 3 starting at column 36
    |> Option.defaultWith (fun _ -> failwith "A unique error message")
```

### Known workarounds

Add message argument

```fsharp
let failIfNone msg (opt: 'a option) : 'a =
    opt |> Option.defaultWith (fun _ -> failwith msg)
```

### Related information

* Windows 10
* 0.21.1
* dotnet 6.0.201

knocte:
Hey @peterhirn thanks for reporting this bug.

Are you sure this is related to Option.defaultWith? The error "Consider using unique error messages with failwith." is supposed to happen when you have more than one failwith call with the same error message. How many failwith calls with "1" do you have in your codebase?

peterhirn:
Hi. I'm not sure if this only happens with `defaultWith` but I ran lint against a single file containing only this function to reproduce the problem. So codebase = this function, error message is unique.

peterhirn:
Rule also applies to every `extern` function, eg.

```
Bad usage of failwith: Consider using unique error messages with failwith.
Error on line 18 starting at column 4
    extern int randombytes_uniform(int upperBound)
    ^                                             
```

knocte:
> Rule also applies to every extern function, eg.

We reproduce the defaultWith issue, but not this one.

peterhirn:
It seems to only happen when there are multiple `extern` functions in a file

```fsharp
open System.Runtime.InteropServices

[<DllImport("libsodium", CallingConvention = CallingConvention.Cdecl)>]
extern int crypto_pwhash(byte [] buffer, int64 bufferLength, byte [] password, int64 passwordLength, byte [] salt, int64 opsLimit, int memLimit, int alg)

[<DllImport("libsodium", CallingConvention = CallingConvention.Cdecl)>]
extern int randombytes_uniform(int upperBound)
```

peterhirn:
I found another problem with this rule when throwing in a if/else

```fsharp
if false then
    ()
else
    failwith "Unique error message"
```

[...]

PR #541, opened 2022-04-04
FailwithBadUsage: fix some false positive and add unit test

janus:
* FailwithBadUsage: new unit test that covers false positive

* FailwithBadUsage: fix false positive

* Fixes: https://github.com/fsprojects/FSharpLint/issues/539

### DIFF
--- a/src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
@@ -8,7 +8,14 @@ open FSharpLint.Framework.Rules
 open System
 open System.Collections.Generic
 
-let mutable failwithErrorMessageList = Set.empty
+type Location =
+    {
+        FileName: string
+        StartLine: int
+        StartColumn: int
+    }
+
+let mutable failwithMessages = Set.empty
 
 type private BadUsageType =
     | EmptyMessage
@@ -43,6 +50,7 @@ let private runner (args: AstNodeRuleParams) =
             | SwallowedException ->
                 "rather use `raise` passing the current exception as innerException (2nd parameter of Exception constructor), otherwise using `failwith` the exception details will be swallowed"
             | NullMessage -> "consider using a non-null error messages as parameter"
+
         let error =
             { Range = range
               Message = String.Format(Resources.GetString "RulesFailwithBadUsage", message)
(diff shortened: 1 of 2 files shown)


src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
```diff
@@ -81,7 +86,7 @@ let private runner (args: AstNodeRuleParams) =
                     range
                     (BadUsageType.EmptyMessage)
                     (None)
-            | SynExpr.Null range -> 
+            | SynExpr.Null range ->
```
knocte, line 89: @janus this line change is not needed

src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
```diff
@@ -62,14 +63,18 @@ let private runner (args: AstNodeRuleParams) =
...
+
+                if testRange then
                     generateError failwithId.idText id range BadUsageType.DuplicateMessage maybeIdentifier
                 else
                     match maybeIdentifier with
                     | Some maybeId ->
                         generateError failwithId.idText id range BadUsageType.SwallowedException (Some maybeId)
                     | _ ->
-                        failwithErrorMessageList <- failwithErrorMessageList.Add(id)
+                        failwithRangeList <- List.append (List.singleton (id, range)) failwithRangeList
```
knocte, line 77: AFAIU you don't need to use append here, you can just concatenate with `::`

knocte:
@aarani can you review please?

aarani:
> @aarani can you review please?

I think by adding range to the list, you are just masking the underlying issue,
why there are multiple items (with the same range) in the list, in the first place.
If there are multiple list items for one failwith that shouldn't be the intended behavior and it should be fixed.

P.S: was it really necessary to change Set to List?

src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
janus: Done

src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
janus: Done

knocte:
Given that @janus and @aarani had some additional discussion about this off-line (well, online, but outside GitHub), I'll update the state of things here:
* Emeka discovered that the reason for the failwith to be inserted twice in the Set is not because of some enumeration/seq problem but because the AST for the testcase shows the failwith twice.
* I then suggested that maybe Emeka's initial solution is not as bad as we thought (adding the range), however, the filename should be added too (range alone is not enough). @aarani still thinks that this is a hacky solution (please feel free to elaborate here Afshin)
* I wonder if the AST containing the failwith twice is an FCS bug? To clear out this theory, one could try the web fantomas tool to check the AST of the testcase (because fantomas updates FCS versions more frequently, I guess they are already using a newer version than FSharpLint).

src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
knocte: @janus you shouldn't say "Done" until you push, I don't see the addressing of this issue in the PR.

PR #565, opened 2022-08-09
Conventions/Naming: rename Accessibility DU to AccessControlLevel

webwarrior-ws:
Also added comment to `AccessControlLevel` describing relationships between values of this DU.


knocte:
@webwarrior-ws looked at the some code after I merged this and realized we should have renamed other stuff, so can you create another PR covering this:

- Parameter "isPublic" in getValueOrFunctionIdents is not of bool type, but AccessControlLevel type, let's rename it.
- Nested function isSynAccessPublic doesn't return bool so it has a confusing name too, let's rename it.
- Function getAccessibility returns AccessControlLevel so let's rename it too.

PR #570, opened 2022-10-14
Rename dotnet-fsharplint to fsharplint-tool

knocte:
Given the previous maintainer seems unresponsive, and is the sole owner of the dotnet-fsharplint nuget package, we may need a way around it (his nuget key expired and we can't renew it without him).

(Longer term, it's maybe better to actually take over the old nuget package called "fsharplint", which is also owned by "fsprojects", but let's do that later.)


knocte:
Matthew replied and now I can create a nuget key myself, so we don't need this anymore.

PR #577, opened 2022-10-21
Adjust to find upcast new redundant

jp-fournier-dev:
Add a case for Update with a new keyword in the expression


knocte:
Good work! Is this PR finished? I ask because it's still marked as DRAFT, but I feel it can already be merged since you added a unit test and your code doesn't break any (old&new) tests.

jp-fournier-dev:
It does add a test but doesn’t really add anything. I’ll make it work properly and then we can merge and fix the issue

Issue #517, opened 2021-10-19
Feature: rule to suggest adding a C#-friendly async overloads


knocte:
### Description

When dealing with F#-async workflows and exposing these APIs as public, we often forget about our C#-dev friends.

### Repro

1. Example a:
```
module Foo =
    let Bar(): Async<unit> =
        Async.Sleep 5
```

2. Example b:
```
type Foo() =
    member __.Bar(): Async<Baz> =
        async { return foobarbaz }
```

### Expected behavior

1. Example a: suggest creation of Foo.BarAsync():
```
module Foo =
    let Bar(): Async<unit> =
        Async.Sleep 5
    let BarAsync(): Task<unit> =
        Bar() |> Async.StartAsTask
```

2. Example b: suggest creation of Foo.BarAsync():
```
type Foo() =
    member __.Bar(): Async<Baz> =
        async { return foobarbaz }
    member this.BarAsync(): Task<Baz> =
        this.Bar() |> Async.StartAsTask
```

PR #533, opened 2021-12-20
Add new rule CSharpFriendlyAsyncOverload

su8898:
Fixes https://github.com/fsprojects/FSharpLint/issues/517

PR #586, opened 2022-11-18
Add new rule C#-FriendlyAsyncOverload

parhamsaremi:
Fixes https://github.com/fsprojects/FSharpLint/issues/517

**NOTE**: this PR supersedes an older [PR](https://github.com/fsprojects/FSharpLint/pull/533)


knocte:
This PR needs to change because we're going to change the convention: instead of having `Foo(): Async<'T>` for the F# method and `FooAsync(): Task<'T>` for the C# one, we're gonna have `AsyncFoo(): Async<'T>`, `FooAsync(): Task<'T>`, and `Foo(): 'T` for the synchronous version. This conclusion comes from a debate between me and Afshin in this PR: https://github.com/nblockchain/NOnion/pull/56

Thorium:
Not all the libraries are meant for C# use. For example type-providers, they just don't work in C#.

knocte:
> Not all the libraries are meant for C# use. For example type-providers, they just don't work in C#.

Good point, that's why this rule won't be enabled by default.

Issue #594, opened 2022-12-02
Rule to recommend not to use the pipe operator if it's only used once


knocte:
### Description

Using the pipe operator is really useful when you want to chain many calls to functions together (using the operator many times). But when the invokation is just one (e.g. param |> func) it becomes a bit pointless (because it's more characters for no real readability gain).

### Testcase that should pass

```
let someFunc someParam =
    someParam
    |> someOtherFunc
    |> yetAnotherFunc
```

### Testcase that should fail

```
let someFunc someParam =
    someParam
    |> someOtherFunc
```

PR #595, opened 2022-12-20
Rule to recommend not to use the pipe operator if it's only used once 

tehraninasab:
Fixes https://github.com/fsprojects/FSharpLint/issues/594

### DIFF
--- a/docs/content/how-tos/rule-configuration.md
+++ b/docs/content/how-tos/rule-configuration.md
@@ -117,4 +117,4 @@ The following rules can be specified for linting.
 - [FavourConsistentThis (FL0074)](rules/FL0074.html)
 - [AvoidTooShortNames (FL0075)](rules/FL0075.html)
 - [FavourStaticEmptyFields (FL0076)](rules/FL0076.html)
-- [AvoidOnePipeOperator (FL0077)](rules/FL0077.html)
+- [AvoidSinglePipeOperator (FL0077)](rules/FL0077.html)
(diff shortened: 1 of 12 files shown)


knocte:
I've realised that this is going too far and the rule is going to raise too many warnings.

Let's have this new test:
```
let someFunc someParam =
    someOtherFunc1 someParam
    |> someOtherFunc2
```

And the rule should not give errors for the above, even if the pipe operator is only used once. Why? Because the alternative would be `someOtherFunc2 (someOtherFunc1 someParam)`, and using the operator in this case is avoiding the use of parentheses, which is fair enough.

BTW let's rename the rule to AvoidOnePipeOperator, to make it crystal clear that multiple pipes are fine.

realmarv:
> I've realised that this is going too far and the rule is going to raise too many warnings.
> 
> Let's have this new test:
> 
> ```
> let someFunc someParam =
>     someOtherFunc1 someParam
>     |> someOtherFunc2
> ```
> 
> And the rule should not give errors for the above, even if the pipe operator is only used once. Why? Because the alternative would be `someOtherFunc2 (someOtherFunc1 someParam)`, and using the operator in this case is avoiding the use of parentheses, which is fair enough.
> 
> BTW let's rename the rule to AvoidOnePipeOperator, to make it crystal clear that multiple pipes are fine.

Added the test and fixed the rule to pass it. But, the name of the rule is already AvoidOnePipeOperator

knocte:
> But, the name of the rule is already AvoidOnePipeOperator

Sorry, I meant `AvoidSinglePipeOperator`.

knocte:
Why is CI red?

Issue #604, opened 2023-10-31
Update to .NET 6 ?


Numpsy:
### Description

Spin off from comments in #603  - 

As it stands:
1) global.json is set to require a .NET 5 SDK to build
2) The command line tool is built as a .NET 5 application.

But - .NET 5 is out of support.
So - 

What are the thoughts on
1) Updating the SDK version used in the build

and/or

2) Retargetting the build to .NET 6 ?

I'm not sure offhand if everything will work as .NET 6 without changes or if anything else would need updating, but I think it is the case that some of the supporting libraries have started dropping support for .NET 5 and needing 6+ so that update might be a pre-requisite of other updates.

baronfel:
I'd advocate for going right to .NET 8, even if this tool keeps targeting the `net6.0` TFM.

Numpsy:
fwiw, I tried building with the .NET 8 RC2 SDK and immediately fell over that issue with ```dotnet tool restore``` being broken, but it's at least only a couple of weeks till the RTM version

knocte:
> I'd advocate for going right to .NET 8, even if this tool keeps targeting the net6.0 TFM.

Sorry, what do you exactly mean with this? Do you mean use .NET8 in CI but have net6 as a value of the <TargetFramework> tag?

Numpsy:
You could possibly make the command line tool .NET 6, but use the version 8 SDK (which could give you the benefits of any other improvements in the build tooling whilst still allowing users to run it just with 6).

Fwiw, I tried building it with the .NET 7 SDK and fell over https://github.com/fsprojects/FAKE/issues/2748 so MSBuild.StructuredLogger probably needs to be updated in the CI builds

knocte:
> You could possibly make the command line tool .NET 6, but use the version 8 SDK for builds (which could give you the
> benefits of any other improvements in the build tooling whilst still allowing users to run it just with 6).

TBH I'm not sure what this means. I guess I need to see a diff to understand it.
But my cautious mentality would be asking to upgrade step by step, first everything to .NET6 in one PR, then we make a release, and then we upgrade to .NET8 like you say here in a subsequent PR, whose changes would land in the next release. This way, in case there are any issues, we can ask users what version they are on, and rollback to a previous version (or update to latest) to see if it fixes their problem.

baronfel:
[...]

PR #603, opened 2023-10-30
Update Paket to v7.2.1

Numpsy:
Paket updated with ```dotnet tool update paket```.

Reasoning: 

The 5.252.0 version it currently uses doesn't understand .NET 6, so it needs to be updated to a newer version.
As the new version still works with .NET 5, it could be updated as a stand alone step to reduce the number of changes needed later.

PR #605, opened 2023-11-01
Build with the .NET 6 SDK

Numpsy:
refs #604

A basic start to update to something supported, and to handle a couple of compiler warnings that occur with newer SDK versions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
         - ubuntu-latest
         - windows-latest
         - macOS-latest
-        dotnet: [5.0.202]
+        dotnet: [5.0.202, 6.0.416]
     runs-on: ${{ matrix.os }}
 
     steps:
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project ToolsVersion="15.0">
   <PropertyGroup>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>
+    <MSBuildTreatWarningsAsErrors>false</MSBuildTreatWarningsAsErrors>
     <IsPackable>false</IsPackable>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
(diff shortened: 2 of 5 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project ToolsVersion="15.0">
   <PropertyGroup>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>
+    <MSBuildTreatWarningsAsErrors>false</MSBuildTreatWarningsAsErrors>
```
Numpsy, line 4: disabled this for now, as there are warnings about .NET 5 being unsupported

src/FSharpLint.Core/Rules/Conventions/RaiseWithTooManyArguments/FailwithBadUsage.fs
```diff
@@ -116,7 +116,7 @@ let private runner (args: AstNodeRuleParams) =
                     | SynPat.Named (_, id, _, _, _) -> checkExpr app (Some id.idText)
                     | _ -> checkExpr app None)
                 | _ -> Array.empty
-        | _ -> Array.empty
+        // | _ -> Array.empty
```
Numpsy, line 119: The newer sdks have this warning here
```
FSharpLint\src\FSharpLint.Core\Rules\Conventions\RaiseWithTooManyArguments\FailwithBadUsage.fs(119,11): error FS0026: This rule will never be matched
```
(which cause a build failure due to WarningsAsErrors being enabled)

tests/FSharpLint.Core.Tests/Rules/Conventions/CyclomaticComplexity.fs
```diff
@@ -229,7 +229,7 @@ let f() =
         let code = $"""Module Program
 let f() = 
     let g() =
-{(makeMatchSnippet (MaxComplexity+1)) |> indent 8} 
+        {(makeMatchSnippet (MaxComplexity+1)) |> indent 8}
```
Numpsy, line 232: The newer SDKs have this warning here
```
FSharpLint\tests\FSharpLint.Core.Tests\Rules\Conventions\CyclomaticComplexity.fs(232,3): error FS0058: Possible incorrect indentation: this token is offside of context started at position (229:9). Try indenting this token further or using standard formatting conventions.
```

.github/workflows/build.yml
```diff
@@ -24,6 +24,10 @@ jobs:
       uses: actions/setup-dotnet@v1
       with:
         dotnet-version: ${{ matrix.dotnet }}
+    - name: Setup .NET 6
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: '6.x'
```
knocte, line 30: @Numpsy why not upgrade from v5 to v6 everywhere? (specifically, am wondering why you left 5.0.x in L15)

Directory.Build.props
knocte: but isn't this PR upgrading from v5 to v6? then this warning should be gone in this PR

Numpsy:
It was an attempt to do one thing at once, and show those two code changes which could be done separately if wanted.

Also it needs Paket to be updated as in #603 as the existing version doesn't understand the newer TFMs

PR #607, opened 2023-11-08
Allow developers to update their Dotnet SDK version

Thorium:
Just a minor update to get this easier to compile for new people.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
   "sdk": {
     "version": "5.0.202"
+    ,"rollForward": "minor"
   }
 }


global.json
```diff
@@ -1,5 +1,6 @@
 {
   "sdk": {
     "version": "5.0.202"
+    ,"rollForward": "minor"
```
bartelink, line 4: How did you pick this, would latestMajor not work/be better?
also suggest not using that formatting

Thorium: Minor is more safe, major typically needs a separate PR for compatibility (see the .NET 6 PRs #605 and #606). Minor forward is just saying that if you have the latest .NET 5 (v.408) installed, you can still compile this repo that is using v.202.

I added the comma intentionally to this line, to not conflict the .NET 6 PR, which is modifying separately the version line above, as we would like to this rolling-minor to be used still after major version update too.

bartelink: Ah, you're way ahead of me; thanks for taking the time to explain in detail!

Numpsy: > Minor is more safe, major typically needs a separate PR for compatibility

yup, I hit a few errors trying to build with the 6 or 7 SDKs that need either code fixes or further changes elsewhere

Issue #341, opened 2019-05-23
Exception thrown when linting file/project through CLI tool


Kelvin4702:
### Description

The following source causes the cli tool to throw an exception when linting the .fs file or the .fsproj that contains the file.
```let first,second = fst e,snd e```

Exception thrown:
```dotnet fsharplint -f foo.fsproj```
```dotnet fsharplint -sf file1.fs```
Succeeds:
```dotnet fsharplint -source "let first,second = fst e,snd e"```

I have not tried the same programatically through FSharpLint.Core.

### Repro steps

1. Create a file with the following source code: ```let first,second = fst e,snd e```.
2. Run the CLI tool on the file or the fsproj containing the file.

### Expected behavior

No exception is thrown.

### Actual behavior

An exception is thrown:
```
An item with the same key has already been added. Key: x
Exception Stack Trace:
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at FSharpLint.Rules.HintMatcher.MatchExpression.matchHintExpr(Arguments arguments) in C:\projects\fsharplint-231\src\FSharpLint.Core\Rules\Hints\HintMatcher.fs:line 253
   at Microsoft.FSharp.Primitives.Basics.List.map2ToFreshConsTail[a,b,c](FSharpList`1 cons, FSharpFunc`3 f, FSharpList`1 xs1, FSharpList`1 xs2)
   at Microsoft.FSharp.Primitives.Basics.List.map2[T1,T2,TResult](FSharpFunc`2 mapping, FSharpList`1 xs1, FSharpList`1 xs2)
   at FSharpLint.Rules.HintMatcher.MatchExpression.doExpressionsMatch(FSharpList`1 expressions, FSharpList`1 hintExpressions, Arguments arguments) in C:\projects\fsharplint-231\src\FSharpLint.Core\Rules\Hints\HintMatcher.fs:line 297
   at Microsoft.FSharp.Primitives.Basics.List.map2ToFreshConsTail[a,b,c](FSharpList`1 cons, FSharpFunc`3 f, FSharpList`1 xs1, FSharpList`1 xs2)
   at Microsoft.FSharp.Primitives.Basics.List.map2[T1,T2,TResult](FSharpFunc`2 mapping, FSharpList`1 xs1, FSharpList`1 xs2)
[...]

PR #608, opened 2023-11-08
Fix for issue #341

Thorium:
The cause of this crash is that there are multiple rules hit for a variable.
Right now what happens: FSharpLint crashes and doesn't produce output.

Instead, if we just ignore the failure of adding a rule failure to a single line of code, the end-user would produce a result.
After fixing the issue the end-user can re-run the lint again and see if the earlier ignored rule still hits or not.

### DIFF
--- a/src/FSharpLint.Core/Rules/Hints/HintMatcher.fs
+++ b/src/FSharpLint.Core/Rules/Hints/HintMatcher.fs
@@ -256,7 +256,7 @@ module private MatchExpression =
             | _ -> NoMatch
         | Expression.Variable(var) ->
             match expr with
-            | AstNode.Expression(expr) -> arguments.MatchedVariables.Add(var, expr)
+            | AstNode.Expression(expr) -> arguments.MatchedVariables.TryAdd(var, expr) |> ignore
             | _ -> ()
             Match([])
         | Expression.Wildcard ->


knocte:
@Thorium hi Tuomas, great contribution! Do you mind adding a regression test?

knocte:
LGTM! Just 2 nits please:
* In the test file, please add EOL before EOF.
* Do you mind changing the order of your commits with interactive rebase? After that, push 1 by 1 (so I can see the 1st one with a red CI, and the 2nd fixing it).

Thorium:
You can test specific commits just by id `git cherry-pick b80db49b7b2f` when in master.
But, it turns out I cannot easily replicate the current runtime crash with an unit-test.
So, I removed the test.

knocte:
> You can test specific commits just by id git cherry-pick b80db49b7b2f when in master.

But I want GitHub to do it for me :) Can you enable GitHubActions in your fork please?

> But, it turns out I cannot easily replicate the current runtime crash with an unit-test.

Mmm, then we should bring some fragment of code from ProvidedTypes.fs like you mentioned in the bug?

Thorium:
I was able already to fix ProvidedTypes.fs with the following workaround:

- I added custom fsharplint.json where I removed so many rules that it didn't hit many duplicate rules at the same code line.
- Then I fixed the issues it found.
- Observe it's now hitting only single rules, so it can run with default config. 

However, as a workaround without this fix, that's a quite un-inviting user experience for new users.

knocte:
> However, as a workaround without this fix, that's a quite un-inviting user experience for new users.

Completely agree, but let's not lose scope here, we're talking about fixing issue #341 here. Can you rescue a previous version of ProvidedTypes.fs that reproduces the bug?

Thorium:
>  Can you rescue a previous version of ProvidedTypes.fs that reproduces the bug?

That's miles along. Rather use the code that is in the original issue. But it's not happening while in source-code (as you can see from the issue), rather when using external files only. Which means it should be an integration test and not a unit test. But I don't see the amount of work to add integration tests would be beneficial on resolving the issue.

knocte:
That's fair enough. I'll then merge this shortly, at least knowing it doesn't break any current tests.

I just need to figure out if I want to merge this before or after next release.

PR #614, opened 2023-11-21
Update the 'setup-dotnet' action in the CI build

Numpsy:
Just an observation that there are several warnings in the CI builds about using old versions of the actions:

image

I don't know how important it actually is, but it does remove a bit of noise, so here it is if its considered useful

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,11 +17,11 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Setup .Net Core
-      uses: actions/setup-dotnet@v1
+      uses: actions/setup-dotnet@v3
       with:
         dotnet-version: '3.1.x'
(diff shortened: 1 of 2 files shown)


knocte:
Just a nitpick: now that we're upgrading, let's upgrade to last version? I think v3 is not the last.

Numpsy:
I was under the impression that setting it to just 'v3' rather than the full version would cause it to use the latest 3.x release? (3.2.0 as it stands)

knocte:
Oh interesting. Then I guess this can go in. That said, pinning to a specific version is always more "stable" in a way, because we would prevent being affected by bugs (regression) that might happen in the latest versions which might leave us scratching our heads.

Numpsy:
I believe that you can set it to a specific version and then set up dependabot to prompt about updates, but I've never tried setting that up myself

Issue #578, opened 2022-10-25
Paket requires extra privileges in Windows


cmeeren:
I am trying to set up a FSharpLint dev environment locally. From a clean repo, I first did `dotnet tool restore`, and then `dotnet fake build`, but the latter fails:

```
There was a problem while setting up the environment:
-> symlink C:\Users\cmeer\.nuget\packages\runtime.native.system.net.security\4.3.1 -> C:\Users\cmeer\Source\Repos\FSharpLint\packages\docs\runtime.native.System.Net.Security failed with error : [1] with output : 
    and error : You do not have sufficient privilege to perform this operation.
```

Do I have to run the build command with administrative privileges? I am hesitant to do that. It should not be needed.

cmeeren:
Seems it's a known issue with paket: https://github.com/fsprojects/Paket/issues/3767#issuecomment-600330976

Do you run with elevated privileges or with Windows in developer mode?

knocte:
Thanks for the bug report and providing a workaround. As for me, I'm using Linux and never got this problem.

We're thinking of dropping paket anyway, now that NuGet has central-package-management capabilities (and also because I've ran into other paket bugs myself).

cmeeren:
That would definitely simplify setup for me. (I mostly just use the IDE, Rider in my case, and let that handle restores, build, and testing. It's always a hurdle to start to contribute to OSS projects where that doesn't work and I have to use build scripts.)

Numpsy:
> Thanks for the bug report and providing a workaround. As for me, I'm using Linux and never got this problem.
> 
> We're thinking of dropping paket anyway, now that NuGet has central-package-management capabilities (and also because I've ran into other paket bugs myself).

What are the current thoughts on Paket vs. Nuget? If we can get the build update to the current .NET SDK then we could try changing it to use directory.packages.props instead?

knocte:
@cmeeren I have the suspicion that PR #615 (which I merged yesterday) might have fixed this issue. Can you re-test and let us know?

cmeeren:
Sorry, I have long since moved on and can not prioritize this now.

knocte:
Ok, then I hope you don't mind that I close this, and if you find the issue again, please reopen!

PR #615, opened 2023-11-22
[RFC] Use 'Packages' storage for the docs when doing a paket restore

Numpsy:
I've been trying to get everything [building with .NET 6](https://github.com/fsprojects/FSharpLint/pull/606) and the related tooling, and have been getting CI build failures in the macOS builds, with errors that look like those described at https://github.com/fsprojects/Paket/issues/3767

It looks like changing Paket to use 'packages' storage instead of 'symlink' avoids that issue, by putting the restored nuget packages directly into the local packages folder instead of symlinking them. (I don't know much about this Paket feature - the docs suggest its for reducing disk usage by avoid duplicates, which might not really be an issue for the builds?)

On a related note, we also have this issue - https://github.com/fsprojects/FSharpLint/issues/578 - where you have to either run the build as admin on windows or set some extra OS options otherwise the package restore will fail. I think this change would avoid that as well.

Note: Changing the paket storage to 'none' helped with package restore issue, but caused the 'docs' build to break for me because the Fornax tool couldn't find the libraries, whereas it works when set to 'packages'

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -29,7 +29,7 @@ group Docs
   source https://api.nuget.org/v3/index.json
 
   framework: netstandard2.0
-  storage: symlink
+  storage: packages
 
   nuget FSharp.Formatting ~> 4.0
   nuget Markdig
--- a/paket.lock
+++ b/paket.lock
@@ -913,7 +913,7 @@ NUGET
       System.Runtime.CompilerServices.Unsafe (>= 4.5.3)
 
 GROUP Docs
-STORAGE: SYMLINK
+STORAGE: PACKAGES
 RESTRICTION: == netstandard2.0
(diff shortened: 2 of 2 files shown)


knocte:
LGTM! Just a nit: can you add your precious PR description text into the commit msg? No reason for them to be different :) (This way, if we move out of github in the future, people would be able to still find the **why** for your commit just by using git blame ;) )

Numpsy:
updated

knocte:
Why did you remove the commit msg title? It's always good to separate the title from the commit msg body by 1 blank line, like if the first line was the subject of an email and the rest the email. And in this case you can just use PR title as commit msg title, and PR description as commit msg body.

PR #619, opened 2023-11-24
Add the BenchmarkDotNet artifacts folder to gitignore

Numpsy:
I was having a go at running the benchmark project, and all the artifacts got listed as changes in Git - and I don't think they should?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,3 +191,6 @@ tests/FSharpLint.FunctionalTest.TestedProject/FSharpLintMSBuildTaskTest/
 DesignTimeBuild/
 .vs/
 .ionide/
+
+# BenchmarkDotNet Results
+tests/FSharpLint.Benchmarks/BenchmarkDotNet.Artifacts


.gitignore
```diff
@@ -190,4 +190,7 @@ tests/FSharpLint.FunctionalTest.TestedProject/FSharpLintMSBuildTaskTest/
 
 DesignTimeBuild/
 .vs/
-.ionide/
\ No newline at end of file
+.ionide/
+
+# BenchmarkDotNet Results
+/tests/FSharpLint.Benchmarks/BenchmarkDotNet.Artifacts
```
knocte, line 196: @Numpsy this being a relative path... shouldn't it **not** start with a slash? otherwise it looks like an absolute path

Numpsy: Hmm, added it via the Visual Studio UI and just took what it did. Updated though.

(fwiw, the ```/paket-files/``` ignore a few lines earlier is the same

PR #626, opened 2023-12-08
CI: merge publish&docs workflow into one

knocte:
Since both of this workflows get executed only when a git tag is pushed, it's better to make them live in the same file, so that:

* They share the same 'on' attribute (there was already an unintended difference between their attributes, caused by a recent change[1]).
* They can depend on each other (now the docs job will only be spawned if the publish one succeeded, so that docs are not updated if there was an issue when running the CI for a certain new tag).

While at it, I renamed these workflows publish.yml and docs.yml into a new merged one 'tagPush.yml' and renamed the other one for the normal build (buildAndTest.yml) to 'branchPush.yml'.

The strategy/matrix elements were also eliminated, since they only make sense in the buildAndTest.yml ones (because we don't want to run the workflow that publishes nuget binaries under different OSs, one is enough).

[1] 64f0d53568e581efbe688a103b58cb06ab31f488


knocte:
Superseded by https://github.com/fsprojects/FSharpLint/pull/635

Issue #627, opened 2023-12-10
Create an API layer for external process (eg: FsAutoComplete)


MrLuje:
### Description

Today, there is no more integration with ionide F# plugin (through FsAutoComplete).
With the current implementation (FSAC directly referencing FSharpLint.Core), it means both FSharpLint & FSAC must share the same version of FSharp.Compiler.Service.

As discussed in https://github.com/fsharp/FsAutoComplete/issues/942, an alternative solution could be to implement a API layer between FSharpLint.Core & FSAC, so FSharp.Compiler.Service could be abstracted using dedicated types (same solution that Fantomas is using)

To summarize the integration with FSAC this new API :
```mermaid
sequenceDiagram
    box green FsAutoComplete process
    participant FsAutoComplete
    participant FSharpLint.Client
    end
    box blue FSharpLint process
    participant FSharpLint.Console
    end
    FsAutoComplete->>FSharpLint.Client: GetDaemon
    FSharpLint.Client->>FSharpLint.Console: Start Daemon
    FSharpLint.Console-->>FSharpLint.Client: 
    FSharpLint.Client->FSharpLint.Console: Setup jsonrpc
    FSharpLint.Console-->>FSharpLint.Client: 
    FSharpLint.Client-->>FsAutoComplete: 
    Note over FsAutoComplete,FSharpLint.Client: Later, when a lint is needed
    FsAutoComplete->>FSharpLint.Client: LintRequest
    FSharpLint.Client->>FSharpLint.Console: LintRequest
    FSharpLint.Console->>FSharpLint.Console: Lint with FSharp.Compiler.Service
    FSharpLint.Console-->>FSharpLint.Client: LintResponse<br/>(no FSharp.Compiler.Service types here)
    FSharpLint.Client-->>FsAutoComplete: LintResponse
```

### TODO:
- [ ] Create the lib layer (Add FSharpLint.Client project)
- [ ] Expose a `version` method for the Client/Console flow
- [ ] Setup the build process to release FSharpLint.Client as a new nuget
- [ ] Expose a `lint` method (that will call [    val lintFile : optionalParams:OptionalLintParameters -> filePath:string -> LintResult](https://github.com/fsprojects/FSharpLint/blob/c7ceca158cbde85891bed4f1427c8502b264a3a2/src/FSharpLint.Core/Application/Lint.fsi#L143) method)

I would like a confirmation that this feature is ok for you guys and I'm willing to work on it

knocte:
[...]

PR #637, opened 2023-12-17
Create an API layer for external process

MrLuje:
## Warning, I branched from the NET 6 version, so https://github.com/fsprojects/FSharpLint/pull/606 should be merged first (ignore the first 4 commits for the review)

WIP on https://github.com/fsprojects/FSharpLint/issues/627, it covers the following points:
- Create the lib layer (Add FSharpLint.Client project)
- Expose a `version` method for the Client/Console flow
- I guess it should also be enough to publish `FSharpLint.Client` nuget

Add `--daemon` flag to FSharpLint.Console
All the "tool location detection" and "process handling" is copied from [Fantomas](https://github.com/fsprojects/fantomas) project and adapted, basically:
- try to locate FSharpLint.Console as a `Local DotnetTool` from current file, then as a `Global DotnetTool` then in `PATH`
- keep a cache of already running daemon (per version)
- communicate between `FSharpLint.Client` and the `FSharpLint.Console` daemon through JsonRpc


knocte:
Good work @MrLuje!

> I'll rebase everything once https://github.com/fsprojects/FSharpLint/pull/606 is merged

Thanks, I plan to re-review and/or merge that PR (or an equivalent) this week, bare with me.

MrLuje:
Added an `FSHARPLINT_SEARCH_PATH_OVERRIDE` env var that will mostly be used for testing that can override search location (global tool & local tool) and a test to ensure FCS is not referenced by FSharpLint.Client

MrLuje:
@knocte done

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
```diff
@@ -0,0 +1,50 @@
+module FSharpLint.Client.LSPFSharpLintServiceTypes
+
+type FSharpLintResponseCode =
+    | ToolNotFound = 1
+    | FileNotFound = 2
+    | FilePathIsNotAbsolute = 3
+    | CancellationWasRequested = 4
+    | DaemonCreationFailed = 5
+    | Version = 6
```
knocte, line 9: @MrLuje what does Version here mean exactly? maybe you mean VersionMismatch ?

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
```diff
@@ -0,0 +1,50 @@
...
+
+type FSharpLintVersion = FSharpLintVersion of string
+
+type FSharpLintExecutableFile = FSharpLintExecutableFile of string
+
+type Folder = Folder of path: string
```
knocte, line 15: @MrLuje how about we use the BCL's type for this, namely DirectoryInfo

tests/FSharpLint.Client.Tests/TestClient.fs
```diff
@@ -0,0 +1,93 @@
+module FSharpLint.Client.Tests
+
+open NUnit.Framework
+open System.IO
+open System
+open Contracts
+open LSPFSharpLintService
+open LSPFSharpLintServiceTypes
+
+let (</>) x y = Path.Combine(x, y)
```
knocte, line 10: @MrLuje cool hack!

src/FSharpLint.Client/FSharpLintToolLocator.fs
```diff
@@ -0,0 +1,256 @@
...
+            )
+        )
+    | ex -> Error(ProcessStartError.UnExpectedException(ps.FileName, ps.Arguments, ex.Message))
+
+let private runToolListCmd (Folder workingDir: Folder) (globalFlag: bool) : Result<string list, DotNetToolListError> =
+    let ps = ProcessStartInfo("dotnet")
+    ps.WorkingDirectory <- workingDir
+
+    if ps.EnvironmentVariables.ContainsKey "DOTNET_CLI_UI_LANGUAGE" then
+        ps.EnvironmentVariables.["DOTNET_CLI_UI_LANGUAGE"] <- "en-us"
```
knocte, line 64: @MrLuje doesn't this line above work in both cases? I mean, if the key was already there and if the key was not (this way, you don't need to call ContainsKey method)

src/FSharpLint.Client/FSharpLintToolLocator.fs
```diff
@@ -0,0 +1,256 @@
...
+    else
+        ps.EnvironmentVariables.Add("DOTNET_CLI_UI_LANGUAGE", "en-us")
+
+    let toolArguments = 
+        Option.ofObj (Environment.GetEnvironmentVariable "FSHARPLINT_SEARCH_PATH_OVERRIDE")
+        |> Option.map(fun env -> $" --tool-path %s{env}")
+        |> Option.defaultValue (if globalFlag then "-g" else String.Empty)
```
knocte, line 71: @MrLuje nit: is there a longer version of this flag? (maybe `--global`?) if so, let's use it, cause short versions of flags should only be used in the terminal (the longer versions are more readable, so ideal for source files)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
MrLuje: No, the opposite. The happy path of the `Version` call
```fsharp
abstract VersionAsync: VersionRequest * ?cancellationToken: CancellationToken -> Task<FSharpLintResponse>
```
Maybe we need a better wording for it ? `DaemonVersion`, `CurrentVersion` or `ToolVersion` ?

knocte: Well, I thought it was an error because all the other enum members are errors. If it's the happy path, shouldn't it map to 0 instead of 6?

src/FSharpLint.Client/FSharpLintToolLocator.fs
MrLuje: Yep, good catch

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
MrLuje: At some point, there will be more, like `LintedFile` for LintRequest so I don't think the order really matter :)

src/FSharpLint.Client/FSharpLintToolLocator.fs
MrLuje: Sure, I'll change it to `--global`

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
MrLuje: I like the fact that when you have a `Folder`, you know it's a valid/existing directory, which you can't be sure with DirectoryInfo (without calling `DirectoryInfo.Exists` before each usage)

Valid `Folder` wasn't really enforced but it can be done with something like this :
```fsharp
type Folder = private Folder of string
    with static member From (folder: string) =
            if System.IO.DirectoryInfo(folder).Exists then
                Folder folder |> Some
            else
                None
```

What do you think ?

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
knocte: Oh I see, I thought these were proc exitCodes! Then, do you mind we make the errors negative and non-errors positive? Something like:

```
type FSharpLintResponseCode =
    | ToolNotFound = -5
    | FileNotFound = -4
    | FilePathIsNotAbsolute = -3
    | CancellationWasRequested = -2
    | DaemonCreationFailed = -1
    | Version = 0
```

Then the next non-error response codes can continue with 1, 2...

knocte: I would even add a prefix to them to make it crystal clear:

```
type FSharpLintResponseCode =
    | ErrToolNotFound = -5
    | ErrFileNotFound = -4
    | FilePathIsNotAbsolute = -3
    | ErrCancellationWasRequested = -2
    | ErrDaemonCreationFailed = -1
    | OkVersion = 0
```

knocte: oh, and renaming Version was also a good idea from you, so `OkCurrentDaemonVersion`? :)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
knocte: I like that! best of both worlds

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
MrLuje: Looks good, go for negative, Ok/Err & `OkCurrentDaemonVersion` 👍

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi
MrLuje: Changes pushed

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
```diff
@@ -0,0 +1,70 @@
...
+    | ErrCancellationWasRequested = -2
+    | ErrDaemonCreationFailed = -1
+    | OkCurrentDaemonVersion = 0
+
+type FSharpLintVersion = FSharpLintVersion of string
+type FSharpLintExecutableFile = FSharpLintExecutableFile of string
+type Folder = private Folder of string 
+with 
+    static member from (filePath: string) =
+        if File.Exists(filePath) then
```
knocte, line 21: @MrLuje should we call IsPathAbsolute here before this check? if it's not, throw an error

knocte: oh wait, I just saw you are calling later Path.GetFullPath, so I'm guessing you're allowing relative paths here, however... (I continue on another line)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
```diff
@@ -0,0 +1,70 @@
...
+    | ErrDaemonCreationFailed = -1
+    | OkCurrentDaemonVersion = 0
+
+type FSharpLintVersion = FSharpLintVersion of string
+type FSharpLintExecutableFile = FSharpLintExecutableFile of string
+type Folder = private Folder of string 
+with 
+    static member from (filePath: string) =
+        if File.Exists(filePath) then 
+            let folder = Path.GetFullPath(filePath) |> Path.GetDirectoryName
```
knocte, line 22: if filePath is /foo/bar/baz, then folder var here is storing `baz` only, which is not what you want

knocte: sorry, `bar` only, not `baz`, but what you should want to extract here is `/foo/bar`

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
```diff
@@ -0,0 +1,70 @@
...
+
+[<RequireQualifiedAccess>]
+type ProcessStartError =
+    | ExecutableFileNotFound of
+        executableFile: string *
+        arguments: string *
+        workingDirectory: string *
+        pathEnvironmentVariable: string *
+        error: string
+    | UnExpectedException of executableFile: string * arguments: string * error: string
```
knocte, line 58: @MrLuje nit, should be Unexpected

src/FSharpLint.Client/Contracts.fsi
```diff
@@ -0,0 +1,28 @@
...
+open System.Threading.Tasks
+
+module Methods =
+
+    [<Literal>]
+    val Version: string = "fsharplint/version"
+
+type VersionRequest =
+    {
+        FilePath: string
```
knocte, line 13: @MrLuje the consistency-chaser in me is wanting to suggest that we create here a similar type to Folder, e.g. `File` which uses FileInfo underneath (instead of DirectoryInfo)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
MrLuje: It's true that an `IsPathAbsolute` call is already made before (https://github.com/MrLuje/FSharpLint/blob/843ef6deccbd1b5f399f572090536dc01c119d96/src/FSharpLint.Client/LSPFSharpLintService.fs#L151) but we could still have paths like `/foo/bar/baz/../../buzz` which is why there is a `Path.GetFullPath` here.
I have some like theses in added tests :)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
MrLuje: You made me doubt but `System.IO.Path.GetFullPath("/foo/bar/baz") |> System.IO.Path.GetDirectoryName` is indeed the expected `/foo/bar` :)

src/FSharpLint.Client/Contracts.fsi
MrLuje: I'm all for consistency but IMHO there is less value for this one :
* VersionRequest is one of the API entrypoint (whereas `Folder` was only used internally) and it may be less intuitive to use
* File check is already covered by `ErrFileNotFound`, so we may want to change it if we go with the `FileInfo` way

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
knocte: Fair enough, but `(System.IO.Path.GetFullPath("/foo/bar/baz") |> System.IO.FileInfo).Directory` is more readable :)

src/FSharpLint.Client/Contracts.fsi
knocte: I know it has less value, but I'm just allergic to the PrimitiveObsession anti-pattern (since I came to know the type `FileInfo`, seeing a `string` for file paths makes my eyes bleed haha). I can make the change myself, I'll push to your branch yeah?

knocte: FTR https://hackernoon.com/what-is-primitive-obsession-and-how-can-we-fix-it-wh2f33ki ;)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
```diff
@@ -0,0 +1,71 @@
...
+
+type FSharpLintVersion = FSharpLintVersion of string
+type FSharpLintExecutableFile = FSharpLintExecutableFile of string
+type Folder = private Folder of string 
+with 
+    static member from (filePath: string) =
+        if File.Exists(filePath) then 
+            // Path.GetFullPath to resolve path like /foo/bar/../baz
+            let folder = ((filePath |> Path.GetFullPath |> FileInfo).Directory).FullName
+            if DirectoryInfo(folder).Exists then
```
knocte, line 24: @MrLuje there's an unnecessary allocation here, the `Directory` property of line 23 is already a DirectoryInfo instance, so you can just use its `Exists` property directly instead of creating another instance.

src/FSharpLint.Client/Contracts.fsi
MrLuje: @knocte sure, go for it :)

src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs
MrLuje: whoops, you're right

src/FSharpLint.Client/Contracts.fsi
knocte: @MrLuje sure, I'll do it this week. BTW can you rebase the PR? looks like there is some conflict

knocte:
@MrLuje side-note, the conflict is actually caused because there is a change in the 1st commit of this PR that has some unnecessary whitespace noise hehe:

> diff --git a/src/FSharpLint.Console/FSharpLint.Console.fsproj b/src/FSharpLint.Console/FSharpLint.Console.fsproj
> index e70053133..f97b5e02b 100644
> --- a/src/FSharpLint.Console/FSharpLint.Console.fsproj
> +++ b/src/FSharpLint.Console/FSharpLint.Console.fsproj
> @@ -1,9 +1,7 @@
> -﻿<Project Sdk="Microsoft.NET.Sdk">
> -
> +<Project Sdk="Microsoft.NET.Sdk">
>    <PropertyGroup>
>      <OutputType>Exe</OutputType>
>      <TargetFrameworks>net6.0</TargetFrameworks>
> -
>      <Title>FSharpLint.Console</Title>
>      <Description>Console application to run FSharpLint.</Description>
>      <PackageTags>F#;fsharp;lint;FSharpLint;fslint;cli</PackageTags>
> @@ -14,15 +12,15 @@
>      <IsPackable>true</IsPackable>
>      <RollForward>Major</RollForward>
>    </PropertyGroup>
> -
>    <ItemGroup>
>      <Compile Include="Output.fs" />
> +    <Compile Include="Version.fs" />
> +    <Compile Include="Daemon.fs" />
>      <Compile Include="Program.fs" />
>    </ItemGroup>
> -
>    <ItemGroup>
>      <ProjectReference Include="..\FSharpLint.Core\FSharpLint.Core.fsproj" />
> +    <ProjectReference Include="..\FSharpLint.Client\FSharpLint.Client.fsproj" />
>    </ItemGroup>
> -
>    <Import Project="..\..\.paket\Paket.Restore.targets" />
> -</Project>
> +</Project>
> \ No newline at end of file

src/FSharpLint.Client/Contracts.fsi
MrLuje: @knocte done

knocte:
> @knocte done

Oh, interesting, CI failed on your fork (macOS job), but I can't see the log because I guess I don't have permissions. Can you share it?

MrLuje:
> > @knocte done
> 
> Oh, interesting, CI failed on your fork (macOS job), but I can't see the log because I guess I don't have permissions. Can you share it?

Not a lot more information from my side (it timed out after 6h, maybe a runner issue ?)
No more issue after a retry
image

knocte:
> @MrLuje sure, I'll do it this week. BTW can you rebase the PR? looks like there is some conflict

Hey @MrLuje, sorry for our delayed action on this, I just pushed a commit that introduces a `File` type, similar to `Folder`.

Now, there are a couple of things to fix here:
1. Early on when you published this PR for the first time, I saw a potential improvement about some lines of code you wrote, but I realised that the "improvable code fragment" could actually be detected by FSharpLint itself, and in fact we had a rule in the works to detect this. This rule is now merged to master, so if you rebase this PR, you will see the offending code and CI will tell you what to change and how :)
2. After working on the `File` type explained above, I've spotted another piece of code that is a bit confusing because it seems to be conflating files with folders, I will go ahead now (after I post this comment) to the "Files changed" tab and point it out there with more context.

src/FSharpLint.Client/LSPFSharpLintService.fs
```diff
@@ -0,0 +1,259 @@
...
+    if requested then
+        Error FSharpLintServiceError.CancellationWasRequested
+    else
+        Ok()
+
+let private getFolderFor filePath (): Result<Folder, FSharpLintServiceError> =
+    let handleFile filePath =
+        if not (isPathAbsolute filePath) then
+            Error FSharpLintServiceError.FilePathIsNotAbsolute
+        else match Folder.from filePath with
```
knocte, line 153: @MrLuje see the code here above? it is feeding a "filePath" to the `Folder` type. So is it a file or a folder? if it's a folder, it should be renamed from filePath to dirPath.

MrLuje: It's a folder (getting the folder of a file)

knocte: @MrLuje ok this is confusing, I thought the point of the `from` function was to convert a path(string) to a proper type. So, in my mind, Folder.from would map a `(folderPath:string)` to a `(folder:Folder)`, and so that's why we included a `File` type equivalent to the `Folder` type that has a `From` method that works this way.

If we want to get the folder where a file lives, I think we should create an API that is more similar to the already existing BCL's System.IO API, to avoid confusions. So, in System.IO, the way to do this is convert the `filePath:string` to a `FileInfo`, and then accessing the `.Directory` property. So how about changing `Folder`'s `From` method to receive a folderPath (not filePath; when given a filePath it should error saying that it is a file, not a folder). So if in this piece of code you want to get the folder where certain path lives, just create it via `File.From` and then let's create a `Folder` property in `File` type?

src/FSharpLint.Client/FSharpLintToolLocator.fs
```diff
@@ -0,0 +1,256 @@
...
+
+    processStart.UseShellExecute <- false
+    processStart.RedirectStandardInput <- true
+    processStart.RedirectStandardOutput <- true
+    processStart.RedirectStandardError <- true
+    processStart.CreateNoWindow <- true
+
+    match startProcess processStart with
+    | Ok daemonProcess ->
+        // fsharplint:disable-next-line RedundantNewKeyword
```
knocte, line 226: @MrLuje why did you disable this rule several times in this PR? did you find a bug?

src/FSharpLint.Client/LSPFSharpLintService.fs
MrLuje: I fear that it might be weird from a discovery POV (signature is a `Folder` but no clue that you have to go through a `File` before)
What about if `Folder.From` can handle both a filePath and a folderPath ? or a `FromFile` / `FromFolder` ?

src/FSharpLint.Client/FSharpLintToolLocator.fs
MrLuje: Maybe ? It is `disposable`, so explicite `new` is recommanded (as a warning)
Without it, I have :
> Usage of `new` keyword here is redundant.F# Linter [FL0014](http://fsprojects.github.io/FSharpLint/how-tos/rules/FL0014.html)

Without the `new` :
> It is recommended that objects supporting the IDisposable interface are created using the syntax 'new Type(args)', rather than 'Type(args)' or 'Type' as a function value representing the constructor, to indicate that resources may be owned by the generated value F# Compiler [760](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/compiler-messages/fs0760)

src/FSharpLint.Client/LSPFSharpLintService.fs
knocte: @MrLuje I like the FromFile/FromFolder idea, let's do it

src/FSharpLint.Client/FSharpLintToolLocator.fs
knocte: @MrLuje hey MrLuje, Mershad is right, see his branch (he edited his commit to add the link) where he removed the 2 comments disabling the rule and CI is still green

MrLuje: Hum, probably my local setup then...
Just to make sure to understand properly, if I go to https://try.fsharp.org/ and type
```fsharp
module Tour.Functions

open System.IO
let s = StreamReader("")
``` 
I get the same warning :
```
It is recommended that objects supporting the IDisposable interface are created using the syntax 'new Type(args)', rather than 'Type(args)' or 'Type' as a function value representing the constructor, to indicate that resources may be owned by the generated value
```
Since `warnings as errors` is enabled in FSharpLint project, I expect the issue I locally have
https://github.com/fsprojects/FSharpLint/blob/08ceae74c1003f4b04ab274f92db758383ce399c/Directory.Build.props#L3-L4
Am i missing something ?

src/FSharpLint.Client/LSPFSharpLintService.fs
MrLuje: Done
I missed it when changing to `FileInfo` but I don't need `Path.GetFullPath` anymore, `FileInfo` handles it properly

src/FSharpLint.Client/FSharpLintToolLocator.fs
knocte: @MrLuje not sure I understand what you're saying. You're bringing up a code snippet that uses a type which is IDisposable, therefore the `new` keyword is needed. That's why you get the warning (from F# compiler).

The rule RedundantNewKeyword is, though, for cases when you're using the `new` keyword in types that are **not** IDisposable. In those cases, using the keyword `new` is not needed. But if you do use it, you don't get a warning from the F# compiler telling you that you don't need it. That's why RedundantNewKeyword exists: to tell you this.

Did I miss something?

knocte: @MrLuje anyway the strange thing here is that you're saying that if you use the `new` keyword, then you get an FSharpLint suggestion, and this is what Mersho refuted above: by showing you a commit whose CI is not red (which means our SelfCheck target passes with no errors; that is, FSharpLint rules are not flagging anything). So please just cherry-pick Mersho's commit from his branch here and push it to your branch.

MrLuje: > @MrLuje not sure I understand what you're saying. You're bringing up a code snippet that uses a type which is IDisposable, therefore the `new` keyword is needed. That's why you get the warning (from F# compiler).
> 
> The rule RedundantNewKeyword is, though, for cases when you're using the `new` keyword in types that are **not** IDisposable. In those cases, using the keyword `new` is not needed. But if you do use it, you don't get a warning from the F# compiler telling you that you don't need it. That's why RedundantNewKeyword exists: to tell you this.
> 
> Did I miss something?

Nope, totally agree with this analysis 👍 
I noticed something else (and I may be missing something about how FSharpLint works) :
* running analysis on the project or solution doesn't trigger it
```bash
$ dotnet src/FSharpLint.Console/bin/Release/net6.0/dotnet-fsharplint.dll lint --file-type project src/FSharpLint.Client/FSharpLint.Client.fsproj
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/obj/Debug/net6.0/.NETCoreApp,Version=v6.0.AssemblyAttributes.fs ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/obj/Debug/net6.0/FSharpLint.Client.AssemblyInfo.fs ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/Contracts.fsi ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/Contracts.fs ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/LSPFSharpLintServiceTypes.fsi ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/LSPFSharpLintServiceTypes.fs ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/FSharpLintToolLocator.fsi ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/FSharpLintToolLocator.fs ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/LSPFSharpLintService.fsi ==========
========== Finished: 0 warnings ==========
========== Linting /home/vince/src/github/mrluje/FSharpLint.worktrees/rw/make_it_all_6/src/FSharpLint.Client/LSPFSharpLintService.fs ==========
========== Finished: 0 warnings ==========
========== Summary: 0 warnings ==========
```

* running it only on FSharpLintToolLocator.fs triggers it
```bash
$ dotnet src/FSharpLint.Console/bin/Release/net6.0/dotnet-fsharplint.dll lint src/FSharpLint.Client/FSharpLintToolLocator.fs 
========== Linting src/FSharpLint.Client/FSharpLintToolLocator.fs ==========
Usage of `new` keyword here is redundant.
Error on line 226 starting at column 22
        let handler = new HeaderDelimitedMessageHandler(
                      ^                                 
See https://fsprojects.github.io/FSharpLint/how-tos/rules/FL0014.html
--------------------------------------------------------------------------------
`x = true` might be able to be refactored into `x`.
Error on line 17 starting at column 53
        if supportedRange.IsSatisfied(parsedVersion, includePrerelease = true) then
                                                     ^                             
See https://fsprojects.github.io/FSharpLint/how-tos/rules/FL0065.html
--------------------------------------------------------------------------------
========== Finished: 2 warnings ==========
========== Summary: 2 warnings ==========
```
In any case, I agree with you there shouldn't be these rules disabled in the final code, I'm pushing Mersho's commit

Numpsy:
So, how much of this was there left to do, besides rebasing on top of all the other changes?
(how much more core work is there to get it over the line?)

MrLuje:
@Numpsy I'll have a look

MrLuje:
@knocte done

src/FSharpLint.Client/FSharpLintToolLocator.fs
```diff
@@ -0,0 +1,253 @@
...
+    let toolArguments =
+        Option.ofObj (Environment.GetEnvironmentVariable "FSHARPLINT_SEARCH_PATH_OVERRIDE")
+        |> Option.map(fun env -> $" --tool-path %s{env}")
+        |> Option.defaultValue (if globalFlag then "--global" else String.Empty)
+
+    ps.CreateNoWindow <- true
+    ps.Arguments <- $"tool list %s{toolArguments}"
+    ps.RedirectStandardOutput <- true
+    ps.RedirectStandardError <- true
+    ps.UseShellExecute <- false
```
knocte: @MrLuje let's use the ProcessStartInfo ctor that accepts all these params? this way you don't need to use the `<-` ugly mutation operator so many times here

Issue #634, opened 2023-12-15
MaxLinesInFunction: counts commented lines


Mersho:
### Description

MaxLinesInFunction also counts commented lines, which is incorrect.

### Steps To Reproduce

• 1.  Make sure MaxLinesInFunction is enabled in fsharplint.json and set the `maxLines` to 2.

• 2. The following files should be included in an example project:

ConsoleApp.fsproj
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
    <WarnOn>3390;$(WarnOn)</WarnOn>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Program.fs" />
  </ItemGroup>

</Project>
```

Program.fs
```
module Program

let dog x =
    // Foo
    // Bar
    // Buzz
    ()
```

• 3. `dotnet fsharplint lint ..\ConsoleApp\ConsoleApp\ConsoleApp.fsproj`

### Current Results

```
========== Linting C:\Users\PC\source\repos\ConsoleApp\ConsoleApp\obj\Debug\net5.0\ConsoleApp.AssemblyInfo.fs ==========
========== Finished: 0 warnings ==========
========== Linting C:\Users\PC\source\repos\ConsoleApp\ConsoleApp\Program.fs ==========
Function was 4 lines long, suggested to be less than 2 lines long.
Error on line 3 starting at column 4
let dog x =
    ^
--------------------------------------------------------------------------------
========== Finished: 1 warnings ==========
========== Summary: 1 warnings ==========
```

### Expected behavior

The rule shouldn't counts commented lines.

### Related information

* Operating system Windows 11
* Branch master
* .NET Runtime, CoreCLR or Mono Version .NET 5

PR #644, opened 2023-12-20
Ignore comments/blank lines when counting number of lines

webwarrior-ws:
Fixes https://github.com/fsprojects/FSharpLint/issues/634

### DIFF
--- a/docs/content/how-tos/rules/FL0062.md
+++ b/docs/content/how-tos/rules/FL0062.md
@@ -24,7 +24,15 @@ Refactor to extract code out into another file.
         "maxLinesInFile": {
             "enabled": false,
             "config": {
-                "maxLinesInFile": 1000
+                "maxLinesInFile": 1000,
+                "skipBlankLines": false,
+                "skipComments": false
             }
         }
     }
+
+*Introduced in `0.23.0`:*
+
+The default setting will be false if the option is not provided in the Rule Settings:
+* *skipBlankLines* - A boolean property that determines whether blank lines will be skipped. (Default false)
+* *skipComments* - A boolean property that determines whether commented lines will be skipped. (Default false)
--- a/docs/content/how-tos/rules/FL0062.md
+++ b/docs/content/how-tos/rules/FL0062.md
@@ -24,15 +24,7 @@ Refactor to extract code out into another file.
         "maxLinesInFile": {
             "enabled": false,
             "config": {
-                "maxLinesInFile": 1000,
-                "skipBlankLines": false,
-                "skipComments": false
+                "maxLinesInFile": 1000
             }
         }
     }
-
-*Introduced in `0.23.0`:*
-
-The default setting will be false if the option is not provided in the Rule Settings:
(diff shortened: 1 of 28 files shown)


knocte:
Two things @webwarrior-ws and/or @Mersho:
1) Please update documentation to reflect the new way to configure the rule (informing about the fact that the attributes skipBlankLines/skipComments were included in v0.23.0 please).
2) What happens if someone configures the rule in the old way? (the JSON that doesn't include skipBlankLines/skipComments?

src/FSharpLint.Core/Rules/Conventions/SourceLength/SourceLengthHelper.fs
```diff
@@ -4,23 +4,90 @@ open System
 open FSharpLint.Framework
 open FSharpLint.Framework.Suggestion
 open FSharp.Compiler.Text
+open FSharpLint.Framework.ExpressionUtilities
+open System.Text.RegularExpressions
```
knocte, line 8: @webwarrior-ws please move System.* to proper place

src/FSharpLint.Core/Rules/Conventions/SourceLength/SourceLengthHelper.fs
```diff
@@ -4,23 +4,90 @@ open System
...
+type Config = 
+    { 
+        MaxLines:int
+        SkipBlankLines: bool
+        SkipComments: bool
+    }
+
+type private MultilineCommentMarker =
+    | Begin of int
+    | End of int
```
knocte, line 20: @webwarrior-ws should be uint?

webwarrior-ws: Since indices in .NET are ints, it's logical to use int here too instead of converting numbers back and forth.

src/FSharpLint.Core/Rules/Conventions/SourceLength/SourceLengthHelper.fs
webwarrior-ws: done

webwarrior-ws:
> 2\. What happens if someone configures the rule in the old way? (the JSON that doesn't include skipBlankLines/skipComments?

These options will be set to `false`

knocte:
> These options will be set to false

Ok, let's add this to the documentation (altho I think @Mersho was doing that?)

knocte:
> (informing about the fact that the attributes skipBlankLines/skipComments were included in v0.23.0 please).

Mersho you missed this ^

docs/content/how-tos/rules/FL0062.md
```diff
@@ -24,7 +24,14 @@ Refactor to extract code out into another file.
...
             "enabled": false,
             "config": {
-                "maxLinesInFile": 1000
+                "maxLinesInFile": 1000,
+                "skipBlankLines": false,
+                "skipComments": false
             }
         }
     }
+*Introduced in `0.23.0`:*
```
knocte, line 33: @Mersho my god man, can you add an EOL before a heading? this is just common sense

PR #645, opened 2023-12-21
AvoidTooShortNames: fix for DU member params & lambda arguments

Mersho:
The rule did not warn about short DU member params.


tests/FSharpLint.Core.Tests/Rules/Conventions/AvoidTooShortNames.fs
```diff
@@ -162,3 +162,15 @@ async {
 } |> Async.RunSynchronously |> ignore<int>"""
 
         Assert.IsTrue this.NoErrorsExist
+
+    [<Test>]
+    member this.AvoidTooShortNamesShouldProduceError_15() =
+        this.Parse """
+type SomeType =
+    | SomeConstructor of int * string * bool
```
knocte, line 170: @Mersho that's not a constructor; SomeType is a DU (maybe let's rename it to SomeDU, and rename SomeConstructor to SomeMember)

PR #648, opened 2023-12-21
Favour nested functions

### LINKED ISSUE #638: New rule FavourNestedFunctions
Testcases:

```
let Foo () =
    ()

let Bar () =
    ()
```
=> No errors

```
let private Foo () =
    ()

let Bar () =
    ()
```
=> No errors

```
let private Foo () =
    ()

let Bar () =
    Foo()
    ()
```
=> Error! (because Foo can be transformed from private to nested)

```
let Bar () =
    let Foo() =
        ()

    Foo()
    ()
```
=> no error (because Foo is nested)

```
let private Foo () =
    ()

let Bar () =
    Foo()
    ()

let Baz () =
    Foo ()
    ()
```
=> no error (because Foo is called by 2 functions: Bar and Baz; therefore, it cannot become nested)

webwarrior-ws:
Add FavourNestedFunctions rule.

Fixes #638


docs/content/how-tos/rule-configuration.md
```diff
@@ -122,3 +122,4 @@ The following rules can be specified for linting.
 - [SuggestUseAutoProperty (FL0079)](rules/FL0079.html)
 - [UnnestedFunctionNames (FL0080)](rules/FL0080.html)
 - [NestedFunctionNames (FL0081)](rules/FL0081.html)
+- [FavourNestedFunctions (FL0082)](rules/FL0082.html)
```
xperiandri, line 125: 0082 is already occupied
```suggestion
- [FavourNestedFunctions (FL0083)](rules/FL0083.html)
```

src/FSharpLint.Core/Rules/Conventions/FavourNestedFunctions.fs
```diff
@@ -0,0 +1,74 @@
...
+    | AstNode.ModuleOrNamespace(SynModuleOrNamespace(_, _, _kind, declarations, _, _, _, _)) ->
+        let privateFunctionIdentifiers = 
+            declarations
+            |> List.toArray
+            |> Array.choose 
+                (fun declaration ->
+                    match declaration with
+                    | FunctionDeclaration(ident, _body, Some(SynAccess.Private)) -> 
+                        Some ident
+                    | _ -> None)
```
xperiandri, line 30:
```suggestion
            |> Seq.choose 
                (fun declaration ->
                    match declaration with
                    | FunctionDeclaration(ident, _body, Some(SynAccess.Private)) -> 
                        Some ident
                    | _ -> None)
            |> Seq.toArray
```

src/FSharpLint.Core/Rules/Conventions/FavourNestedFunctions.fs
```diff
@@ -0,0 +1,74 @@
...
+            |> Array.choose
+                (fun currFunctionIdentifier ->
+                    match ExpressionUtilities.getSymbolFromIdent args.CheckInfo (SynExpr.Ident currFunctionIdentifier) with
+                    | Some symbolUse ->
+                        let numberOfOtherFunctionsCurrFunctionIsUsedIn =
+                            otherFunctionBodies
+                            |> List.filter (fun funcBody -> 
+                                checkInfo.GetUsesOfSymbolInFile symbolUse.Symbol
+                                |> Array.exists (fun usage -> ExpressionUtilities.rangeContainsOtherRange funcBody.Range usage.Range))
+                            |> List.length
```
xperiandri, line 54:
```suggestion
                            |> Seq.filter (fun funcBody -> 
                                checkInfo.GetUsesOfSymbolInFile symbolUse.Symbol
                                |> Array.exists (fun usage -> ExpressionUtilities.rangeContainsOtherRange funcBody.Range usage.Range))
                            |> Seq.length
```

xperiandri (changes requested):
Looks good, just a few remarks

xperiandri:
Could you rebase and fix?

docs/content/how-tos/rules/FL0091.md
```diff
@@ -0,0 +1,31 @@
+---
+title: FL0091
+category: how-to
+hide_menu: true
+---
+
+# FavourNestedFunctions (FL0091)
+
+*Introduced in `0.26.11`*
```
knocte: @webwarrior-ws this version doesn't and will never exist, use .12

knocte:
> Literals can't be nested inside functions. Only module-level.

Then that's a false positive to fix before merging. Anyway, my point still stands wrt module elements that are not functions, just fields.

knocte:
> So maybe don't apply the rule if the function has any attributes?

Right, let's add a test for it.

Issue #650, opened 2023-12-25
New rule AvoidMisleadingRecursiveKeywordInNonRecursiveFuncs


knocte:
Testcases:

Should error:
```
let rec Foo someParam =
    ()
```

Should not error:
```
let rec Foo someParam =
    if someParam then
        Foo()
    else
        ()
```

PR #652, opened 2023-12-26
New rule UnneededRecKeyword

Mersho:
Fixes https://github.com/fsprojects/FSharpLint/issues/650


docs/content/how-tos/rules/FL0082.md
```diff
@@ -0,0 +1,29 @@
+---
+title: FL0082
+category: how-to
+hide_menu: true
+---
+
+# AvoidMisleadingRecursiveKeywordInNonRecursiveFuncs (FL0082)
+
+*Introduced in `0.21.9`*
```
knocte, line 9: @Mersho huh?

knocte:
@Mersho please rebase this, and after that add a commit that renames the rule to UnneededRecKeyword

knocte:
@webwarrior-ws please review

docs/content/how-tos/rules/FL0083.md
```diff
@@ -0,0 +1,29 @@
...
+hide_menu: true
+---
+
+# UnneededRecKeyword (FL0083)
+
+*Introduced in `0.23.1`*
+
+## Cause
+
+Recursive functions (functions with a "rec" keyword) does not call themselves.
```
webwarrior-ws: Grammatical error. Either use plural and "do", or use singular ("function", "itself "not "function", "themselves")

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
```diff
@@ -0,0 +1,44 @@
...
+open FSharpLint.Framework
+open FSharpLint.Framework.Suggestion
+
+let runner (args: AstNodeRuleParams) =
+    match args.AstNode, args.CheckInfo with
+    | AstNode.ModuleDeclaration (SynModuleDecl.Let (isRecursive, bindings, _)), Some checkInfo when isRecursive ->
+        match bindings with
+        | SynBinding (_, _, _, _, _, _, _, SynPat.LongIdent (longDotId, _, _, _, _, range), _, _, _, _) :: _ ->
+            let symbolUses = checkInfo.GetAllUsesOfAllSymbolsInFile()
+            let funcName = longDotId.Lid.Head.idText
```
webwarrior-ws, line 16: If you expect long identifier to have exactly 1 element, it's better to explicitly pattern match it (`SynLongIdent([ident], _)` instead of `longDotId` in pattern)

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
```diff
@@ -0,0 +1,44 @@
...
+
+            let functionCalls =
+                symbolUses
+                |> Seq.filter (fun (symbol) -> symbol.Symbol.DisplayName = funcName)
```
webwarrior-ws, line 20: You also need to check if symbol is used inside function body. You can compare ranges for that.

docs/content/how-tos/rules/FL0083.md
webwarrior-ws: Nevermind, it got fixed since I started a review

knocte: Yep thanks, I added a commit addressing this.

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
knocte: @webwarrior-ws isn't it better to add a new test first to cover this scenario?

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
knocte: @webwarrior-ws would we need a new test to cover this?

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
webwarrior-ws: Yes, a test with function used outside function body some would be helpful

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
webwarrior-ws: I don't think you can define a function with dotted identifier. So there should always be 1 element in idents list.

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
knocte: @Mersho ok Mershad go ahead and add this, with 2 commits like we explained above

src/FSharpLint.Core/Rules/Conventions/UnneededRecKeyword.fs
knocte: not sure what you mean but yeah please Mershad address this (and discuss in telegram with Taras if needed)

Mersho: > @Mersho why did you mark this resolved? I don't see a commit addressing it

we discussed about it and there is no failure test I can create for it.

knocte: that doesn't excuse anything, where is the commit addressing it? (even if it doesn't add a test)

Mersho: It has been addressed in this commit:

https://github.com/fsprojects/FSharpLint/pull/652/commits/aefd4fbd46ef6e97177db7d72f6bbefa5e9962db

i can amend and explain why long identifier have exactly 1 element.

PR #656, opened 2023-12-28
Add quick fix to rule82

webwarrior-ws:
Implement quick fix for `UsedUnderscorePrefixedElements` rule.
It renames identifier with `_` prefix to identifier without this prefix (e.g. `_foo` -> `foo`).
Quick fix is only applied if no other identifiers with suggested name (without underscore) exist.
There are cases when such identifiers exist, but not in the same scope, and still no quick fix is applied. That is because checking the scope proved to be much more difficult than expected.


tests/FSharpLint.Core.Tests/Rules/TestRuleBase.fs
```diff
@@ -75,24 +75,27 @@ type TestRuleBase () =
...
-                    .Remove(startIndex, endIndex - startIndex)
-                    .Insert(startIndex, fix.ToText)
-                    .ToString()
-            | _ -> source
-        | None -> source
+
+        Seq.fold 
+            (fun (source: string) (lazyFix: Lazy<Option<SuggestedFix>>) ->
+                match lazyFix.Value with
+                | Some(fix) ->
```
xperiandri, line 85:
```suggestion
                | Some fix ->
```

tests/FSharpLint.Core.Tests/Rules/TestRuleBase.fs
```diff
@@ -75,24 +75,27 @@ type TestRuleBase () =
...
+                    let startIndex = ExpressionUtilities.findPos fix.FromRange.Start source
+                    let endIndex = ExpressionUtilities.findPos fix.FromRange.End source
+
+                    match (startIndex, endIndex) with
+                    | (Some(startIndex), Some(endIndex)) ->
```
xperiandri, line 90:
```suggestion
                    | (Some startIndex, Some endIndex) ->
```

xperiandri (changes requested):
I would prefer such syntax, otherwise good

knocte:
Let's squash the last commit properly.

webwarrior-ws:
> Let's squash the last commit properly.

Squashed.

PR #660, opened 2023-12-30
README.md,GitHubCI: add nice 'make' experience

knocte:
In Unix (macOS and Linux), it's just the simplest Makefile that calls FAKE underneath.
In Windows, it's a batch file that does the same.

### DIFF
--- a/make.cmd
+++ b/make.cmd
@@ -1,0 +1,23 @@
+@ECHO OFF
+where /q dotnet
+IF ERRORLEVEL 1 (
+    ECHO "ERROR: 'dotnet' not found. Please ensure you have installed .NET (the version specified in global.json)" && EXIT /b 1
+) ELSE (
+    IF "%~1" == "" (
+        dotnet fsi build.fsx --target Build
+    ) ELSE (
+        IF "%~1" == "check" (
+            dotnet fsi build.fsx --target Test
+        ) ELSE (
+            IF "%~1" == "selfcheck" (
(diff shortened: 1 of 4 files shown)


Numpsy:
MAybe it's still worth finishing this one? (see https://github.com/fsprojects/FSharpLint/pull/734 about updating the docs to not mention fake-cli any more)

make.bat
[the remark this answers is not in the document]
knocte: can do, but why? I'm curious

make.cmd
```diff
@@ -0,0 +1,23 @@
...
+            ) ELSE (
+                IF "%~1" == "docs" (
+                    dotnet fsi build.fsx --target Docs
+                ) ELSE (
+                    ECHO "Target was not recognized" && EXIT /b 1
+                )
+            )
+        )
+    )
+)
```
xperiandri, line 23: Shouldn't we just pass the argument as is `-t %~1` without this long match?

knocte: I thought about this, but there's one target that doesn't match: check maps to Test. And I decided to make it `make check` because it is a convention to call the unit tests target this way in the Makefile world.

Numpsy:
Has this change upset the documentation generation in the CI?

I'm getting a 404 error from https://fsprojects.github.io/FSharpLint/ and the build log for the last CI build contains

image

which makes it look like the documentation isn't actually being built in the CI builds.

PR #663, opened 2023-12-31
Add LintFile method to API layer

MrLuje:
## Branched from #637, so it should be merged first

See https://github.com/mrluje/FSharpLint/compare/9635b9a27a8915d82a94e8eeb7931e9931e1f2b3...MrLuje:FSharpLint:lint_layer diff for relevant changes

WIP on #627:
* Expose a lint method (that will call [val lintFile : optionalParams:OptionalLintParameters -> filePath:string -> LintResult](https://github.com/fsprojects/FSharpLint/blob/c7ceca158cbde85891bed4f1427c8502b264a3a2/src/FSharpLint.Core/Application/Lint.fsi#L143) method)

For reminder, the purpose of the API layer is to abstract away all FCS types so the caller can have a different FCS version than FSharpLint.

I duplicated types from https://github.com/fsprojects/FSharpLint/blob/master/src/FSharpLint.Core/Framework/Suggestion.fs to `FSharpLint.Client` and prefixed them with `Client` (eg: `SuggestedFix` => `ClientSuggestedFix`) to tell them apart. (I created `ClientRange` from scratch to avoid the FCS dependency) but that's open to discussion.

```fsharp
type LintFileRequest =
    {
        FilePath: string
        LintConfigPath: string option
    }
```
For now, the `LintFileRequest` takes the path to the file to lint and an optional `LintConfigPath` to a `fsharplint.json` file but I'm wondering about other scenario :
* If no config is specified, `defaultConfiguration` will be used (same with an invalid `LintConfigPath`)
* If a valid `fsharplint.json` is specified through `LintConfigPath`, it will be used
* What about if `LintConfigPath` isn't specified but there is an existing `fsharplint.json` in the project ? is there a standard location for `fsharplint.json` or an existing detection method ? (next to solution file, parse [fc]sproj, etc)


build.fsx
```diff
@@ -309,7 +309,7 @@ Target.create "Release" DoNothing
 
 "Clean"
   ==> "Build"
-  ==> "Test"
+//   ==> "Test"
```
knocte: @MrLuje let's reenable this now?

PR #667, opened 2024-01-05
Fix recent regression in MaxLinesIn* rules

Mersho:
PR #644 accidentally introduced a bug that would render the MaxLinesIn* rules useless.

### DIFF
--- a/tests/FSharpLint.Core.Tests/Rules/Conventions/SourceLength.fs
+++ b/tests/FSharpLint.Core.Tests/Rules/Conventions/SourceLength.fs
@@ -1,10 +1,25 @@
 module FSharpLint.Core.Tests.Rules.Conventions.SourceLength
 
+open System
 open NUnit.Framework
 open FSharpLint.Rules
 open FSharpLint.Rules.Helper.SourceLength
 
-let generateNewLines numNewLines = Array.create numNewLines "\n" |> String.concat ""
+let generateNewLines numNewLines numIndents =
+    Array.mapi
+        (fun index _ ->
+            let indentationChars =
+                if index = 0 then
+                    String.Empty
+                else
+                    String.replicate numIndents " "
+            sprintf "%sprintf System.String.Empty\n" indentationChars)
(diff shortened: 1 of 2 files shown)


tests/FSharpLint.Core.Tests/Rules/Conventions/SourceLength.fs
```diff
@@ -4,7 +4,13 @@ open NUnit.Framework
 open FSharpLint.Rules
 open FSharpLint.Rules.Helper.SourceLength
 
-let generateNewLines numNewLines = Array.create numNewLines "\n" |> String.concat ""
+let generateNewLines numNewLines numIndents =
+    Array.create numNewLines "\n"
+    |> String.concat (sprintf "%sprintf \"\"" (String.replicate numIndents " "))
+
+let generateAbstractMembers numMembers numIndents =
+    Array.init numMembers (fun index -> sprintf "abstract member Foo%d : unit -> unit\n" index)
```
knocte, line 12: @Mersho our F# guidelines say you have to use %i for integers, why do you keep using d if i->integer is easier to remember? what does d mean to you?

tests/FSharpLint.Core.Tests/Rules/Conventions/SourceLength.fs
```diff
@@ -230,8 +236,8 @@ type TestMaxLinesInClass() =
         this.Parse(sprintf """
 module Program
   type MyClass2() as this =
-    %s
-    member this.PrintMessage() = ()""" (generateNewLines ClassLength))
+    do%s
```
knocte, line 239: @Mersho please put the %s on the next line and make generateNewLines NOT start with an EOL

tests/FSharpLint.Core.Tests/Rules/Conventions/SourceLength.fs
Mersho: I think d stands for digit, it's a habit from other programming languages (such as CPP).

PR #673, opened 2024-01-08
Add EnsureTailCallDiagnosticsInRecursiveFunctions rule

### LINKED ISSUE #651: New rule: EnsureTailCallDiagnosticsInRecursiveFunctions
### Description

Context: https://stackoverflow.com/questions/806712/how-do-i-know-if-a-function-is-tail-recursive-in-f/77532717#77532717

### Testcases

Should error:
```
let rec Foo someParam =
    if someParam then
        Foo()
    else
        ()
```

Should not error:
```
[<TailCall>]
let rec Foo someParam =
    if someParam then
        Foo()
    else
        ()
```

Should error (but with different error message):
```
let rec Foo someParam =
    ()
```

In the latter testcase, error message should be same as this rule's err msg: https://github.com/fsprojects/FSharpLint/issues/650 (please don't violate DRY). Otherwise, the error message should advise both about adding the TailCall attribute **and** `<WarningsAsErrors>FS3569</WarningsAsErrors>` (NOTE: it should also mention that this feature requires .NETv8.0 or newer).

webwarrior-ws:
Add EnsureTailCallDiagnosticsInRecursiveFunctions rule and tests for it.

Fixes #651

### DIFF
--- a/src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
@@ -36,7 +36,7 @@ let runner (args: AstNodeRuleParams) =
             else
                 emitWarning func |> Array.singleton
         else
-            UnneededRecKeyword.emitWarning func |> Array.singleton
+            Array.empty
     | _ -> Array.empty
 
 let rule =
--- a/src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
@@ -1,0 +1,20 @@
+﻿module FSharpLint.Rules.EnsureTailCallDiagnosticsInRecursiveFunctions
+
+open System
+
+open FSharpLint.Framework
+open FSharpLint.Framework.Suggestion
+open FSharp.Compiler.Syntax
+open FSharp.Compiler.Text
+open FSharp.Compiler.CodeAnalysis
+open FSharpLint.Framework.Ast
+open FSharpLint.Framework.Rules
+
+let runner (args: AstNodeRuleParams) =
+    failwith "Not yet implemented"
+
+let rule =
+    { Name = "EnsureTailCallDiagnosticsInRecursiveFunctions"
+      Identifier = Identifiers.EnsureTailCallDiagnosticsInRecursiveFunctions
+      RuleConfig = { AstNodeRuleConfig.Runner = runner; Cleanup = ignore } }
+    |> AstNodeRule
--- a/src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
@@ -5,57 +5,38 @@ open System
 open FSharpLint.Framework
 open FSharpLint.Framework.Suggestion
 open FSharp.Compiler.Syntax
-open FSharp.Compiler.Text
-open FSharp.Compiler.CodeAnalysis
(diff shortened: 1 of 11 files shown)


src/FSharpLint.Core/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
```diff
@@ -0,0 +1,65 @@
...
+                            funcName
+                        )
+                      SuggestedFix = None
+                      TypeChecks = list.Empty }
+                    |> Array.singleton
+            else
+                { Range = range
+                  Message =
+                    String.Format(
+                        Resources.GetString "RulesUnneededRecKeyword",
```
knocte, line 52: @webwarrior-ws it's good that this rule doesn't copy a msg from another rule, but when I requested to respect DRY I actually meant it on a deeper level: this rule should call a function from the other rule, not repeat its logic here

webwarrior-ws: There is not much code to reuse and extracting common functionality to use in both rules will likely lead to more complexity.

knocte: @webwarrior-ws then do this in a new commit and after I see the diff of it I will decide if we drop it or we keep it

knocte:
Good work so far @webwarrior-ws . Now, in a previous commit that you had pushed, I saw that the rule UnneededRecKeyword was firing for some code that had some `and` keywords, can you create a different PR covering that scenario please?

tests/FSharpLint.Core.Tests/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
```diff
@@ -0,0 +1,47 @@
...
+        this.Parse """
+let rec Foo someParam =
+    ()
+
+module Bar =    
+    let Foo = 0"""
+
+        Assert.IsTrue 
+            <| this.ErrorWithMessageExistsAt(
+                """The 'Foo' function has a "rec" keyword, but it is not really recursive, consider removing it.""",
```
knocte, line 45: @webwarrior-ws I've changed my mind about this test, essentially because if I merge PR#666 then it will give misleading info. So let's change this test to Assert.NoErrorsExist

knocte: (but don't undo the refactoring commit please, I think it's good.)

knocte:
@webwarrior-ws can you rebase this please

docs/content/how-tos/rules/FL0084.md
```diff
@@ -0,0 +1,32 @@
+---
+title: FL0084
+category: how-to
+hide_menu: true
+---
+
+# EnsureTailCallDiagnosticsInRecursiveFunctions (FL0084)
+
+*Introduced in `0.23.8`*
```
knocte, line 9: @webwarrior-ws let's change this to 0.24.2

webwarrior-ws: done

webwarrior-ws:
> @webwarrior-ws can you rebase this please

rebased

tests/FSharpLint.Core.Tests/Rules/Conventions/EnsureTailCallDiagnosticsInRecursiveFunctions.fs
```diff
@@ -0,0 +1,43 @@
...
+        Assert.IsTrue this.NoErrorsExist
+
+    [<Test>]
+    member this.``Should not error when function is defined with rec keyword but is not recursive``() =
+        this.Parse """
+let rec Foo someParam =
+    ()
+
+module Bar =    
+    let Foo = 0"""
```
knocte, line 41: @webwarrior-ws please place all triple-quote closing elements `"""` at next line

PR #680, opened 2024-01-11
DRAFT: paket move to FAKE-only

webwarrior-ws:
Move from paket to Central Package Management.


knocte:
@webwarrior-ws can you try removing NU1605 now please?

webwarrior-ws:
> @webwarrior-ws can you try removing NU1605 now please?

@knocte where is it?

knocte:
> @knocte where is it?

https://github.com/fsprojects/FSharpLint/blob/58dd6b6704f2787c8c1dfdfa838ba28e29847dc5/Directory.Build.props#L9

You were lucky that I was not afk now :) but please, next time, for obvious questions use grep ;)

webwarrior-ws:
Build fails with multiple errors if I remove NU1605 from `<NoWarn>`

knocte:
If the last commit worked, then it shouldn't be marked as WIP anymore.

BTW when writing commit messages, if you are explaining 2 things that the commit does, please list them with bullets. Otherwise just listing each in a sentence, in the same paragraph, simply separated with a dot (with no way to join the sentences such as "hence", or something) feels kinda strange.

knocte:
Good work wrt "build: re-enable NU1605" commit. Now, looking at it, it looks fairly simple. Are you sure we cannot separate into its own PR (aka: do it in master while it still paket-based)?

webwarrior-ws:
> Good work wrt "build: re-enable NU1605" commit. Now, looking at it, it looks fairly simple. Are you sure we cannot separate into its own PR (aka: do it in master while it still paket-based)?

I tried, and it didn't work. Microsoft.Build.* versions are set by Visual Studio and cause package downgrades.

knocte:
> I tried, and it didn't work. Microsoft.Build.* versions are set by Visual Studio and cause package downgrades.

I never mentioned VS here. The topic is paket vs CPM (at CI level).

webwarrior-ws:
> > I tried, and it didn't work. Microsoft.Build.* versions are set by Visual Studio and cause package downgrades.
> 
> I never mentioned VS here. The topic is paket vs CPM (at CI level).

NU1605 was disabled because build in Visual Studio failed. See 58dd6b6

knocte:
> NU1605 was disabled because build in Visual Studio failed.

If you remove that NoWarn in master, then CI doesn't build. So we're talking about CI here, not VS.

webwarrior-ws:
CI failure seems to be a random bug - CI run for the same commit in [cpm-rebased-2](https://github.com/webwarrior-ws/FSharpLint/tree/cpm-rebased-2) branch is successful: https://github.com/webwarrior-ws/FSharpLint/actions/runs/7828237116

src/FSharpLint.Console/FSharpLint.Console.fsproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
Numpsy, line 6: Is this project file supposed to have both ```TargetFramework``` and ```TargetFrameworks```

knocte: Good catch, @webwarrior-ws please fix

webwarrior-ws: fixed

xperiandri:
I think so, @webwarrior-ws what do you think?

knocte:
It seems master branch migrated out of Paket thanks to one of the commits in PR https://github.com/fsprojects/FSharpLint/pull/722 . But I think what's missing in master that this PR included is Central-Package-Management. Can you confirm @webwarrior-ws ?

knocte:
Good point thanks! Great work everyone.

PR #681, opened 2024-01-16
Console: enhanced warning reporting

Mersho:
This feature is particularly useful in environments like VSCode, which highlights these locations and makes them clickable. When clicked, VSCode will open the corresponding file at the exact line and position of the cursor. This functionality allows developers to quickly navigate to the source of the issue, significantly speeding up the debugging process.


Mersho:
Also we could use workflow annotations, but it would cause extra warnings. This is because the "SelfCheck" step runs three times, once for each operating system and we may encounter additional and repetitive warnings.

knocte:
>  This is because the "SelfCheck" step runs three times, once for each operating system and we may encounter additional and repetitive warnings.

I don't understand this claim.

src/FSharpLint.Console/Output.fs
```diff
@@ -14,18 +14,24 @@ type IOutput =
...
-        String.Format(error, range.StartLine, range.StartColumn)
-
-    let highlightErrorText (range:Range) (errorLine:string) =
+        let startLine = range.StartLine
+        let startColumn = range.StartColumn
+        String.Format(
+            error,
+            startLine,
+            startColumn,
+            sprintf "%s(%i,%i)" filePath startLine startColumn
```
xperiandri, line 25:
```suggestion
            $"%s{filePath}(%i{startLine},%i{startColumn})"
```

src/FSharpLint.Console/Output.fs
```diff
@@ -14,18 +14,24 @@ type IOutput =
...
+        )
+    let highlightErrorText (filePath: string) (range:Range) (errorLine:string) =
         let highlightColumnLine =
             if String.length errorLine = 0 then "^"
             else
                 errorLine
                 |> Seq.mapi (fun i _ -> if i = range.StartColumn then "^" else " ")
                 |> Seq.reduce (+)
-        getErrorMessage range + Environment.NewLine + errorLine + Environment.NewLine + highlightColumnLine
+        getErrorMessage range filePath + Environment.NewLine + errorLine + Environment.NewLine + highlightColumnLine
```
xperiandri, line 34: use string interpolation

xperiandri (commented):
Rebase, please

Mersho:
I rebased, and here is what the new warning message looks like:

```
$ dotnet run --project .\src\FSharpLint.Console\ --framework net9.0 lint C:\Users\PC\source\repos\FSharpLint\tests\FSharpLint.Benchmarks\Benchmark.fs
========== Linting C:\Users\PC\source\repos\FSharpLint\tests\FSharpLint.Benchmarks\Benchmark.fs ==========
`x = true` might be able to be refactored into `x`.
Error on line 14 starting at column 43: C:\Users\PC\source\repos\FSharpLint\tests\FSharpLint.Benchmarks\Benchmark.fs(14,43)
        let checker = FSharpChecker.Create(keepAssemblyContents=true)
                                           ^
See https://fsprojects.github.io/FSharpLint/how-tos/rules/FL0065.html
--------------------------------------------------------------------------------
========== Finished: 1 warnings ==========
========== Summary: 1 warnings ==========
```

xperiandri:
But at the beginning of the output, there is a line `========== Linting C:\Users\PC\source\repos\FSharpLint\tests\FSharpLint.Benchmarks\Benchmark.fs ==========`
And MSBuild output format switch already defines a very similar format.
I don't think that this change makes sense

Could you compare your new output to the output with `-f msbuild` switch and check if it meets your needs?

knocte:
> I don't think that this change makes sense

Agreed, sorry.

PR #688, opened 2024-01-22
paket: upgrade Ionide.ProjInfo to 0.61.3

webwarrior-ws:
Upgrade Ionide.ProjInfo to version 0.61.3 so that more project files can be processed by FSharpLint.

`dotnet paket update --keep-patch` command was used to update paket.lock file.


knocte:
@webwarrior-ws why didn't you upgrade to 0.62.0?

webwarrior-ws:
> @webwarrior-ws why didn't you upgrade to 0.62.0?

It depends on FSharp.Core >= 7, while we are using version 6.

PR #692, opened 2024-01-26
docs: improve/elaborate main project description

knocte:
By just describing as 'style checking tool', one could have thought
of FSharpLint as a formatting tool (sort of an alternative to fantomas)
but we cover way more than that. And to give a better idea of what
kind of issues can be detected, we put three different kinds of rules
here.


knocte:
@xperiandri thanks for your improvements, included them and merged; is English your mother tongue? Mine isn't :sob:

PR #657, opened 2023-12-28
Fix error linting projects that use .NET Framework

webwarrior-ws:
* Upgrade Ionide.ProjInfo to 0.58.0 (needed to bring the fix for parsing legacy projects, upstream PR: https://github.com/ionide/proj-info/pull/131 , upstream commit: https://github.com/ionide/proj-info/commit/3542cee96562b1e64c20a6b88278b1a8346d83f3 ).
* Upgrade FSharp.Compiler.Service to version 41 (needed because the new Ionide.ProjInfo version, depends on a higher version of F.C.S.).
* Made changes to framework, rules and test framework code necessary to accomodate API changes in this version 41.x of FCS compared to version 40.

Fixes https://github.com/fsprojects/FSharpLint/issues/336

Issue #620, opened 2023-11-24
What's the situation with the Benchmarks project?


Numpsy:
Under the 'tests' folder there is a 'FSharpLint.Benchmarks' project, that uses BenchmarkDotNet.

As it stands, that project targets .NET 5.0:
https://github.com/fsprojects/FSharpLint/blob/598788db86dc9c65f9c14c3ecc694f9708d0b930/tests/FSharpLint.Benchmarks/FSharpLint.Benchmarks.fsproj#L5

However, the benchmark code itself it set to try to run the benchmarks as .NET Core 2.1,
https://github.com/fsprojects/FSharpLint/blob/598788db86dc9c65f9c14c3ecc694f9708d0b930/tests/FSharpLint.Benchmarks/Program.fs#L13

So when I try to run it I get errors like
```
error NU1201: Project FSharpLint.Benchmarks is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Project FSharpLint.Benchmarks supports: net5.0 (.NETCoreApp,Version=v5.0)
```

In addition, if I try to change it to test as .NET 5 (or 6 with the in-progress 6 update) then I instead get
```
Please install the latest Microsoft.Diagnostics.Tracing.TraceEvent package in the project with benchmarks so MSBuild can copy the native dependencies of TraceEvent to the output folder.
```

The version of BenchmarkDotNet it's using is rather old and could do with an update (and if it's done after the .NET 6 change then it can use the .NET 6 version rather than the .NET Standard version, which removes a few dependencies), but - given that it seems to be not presently functional I thought I'd check before trying a PR for an update

knocte:
I've never worked on this project, so I don't know much about it sorry. But if you propose PRs that fixes it, they are welcome! (And I guess we should include it in CI to make sure it doesn't rot again?)

PR #694, opened 2024-02-05
Run the benchmarks using the default runtime rather than trying to fo…

Numpsy:
…rce .NET Core 2.1 (as that won't work when the project is built as .NET 6.0)

refs #620 

It looks like BenchmarkDotNet was updated in #657 and that has fixed the ```Microsoft.Diagnostics.Tracing.TraceEvent``` related error, so now it just needs updtaing to not try running as .NET Core 2.1 - if left at default it will run as .NET 6 to match the project settings.

It could alternatively be changed to use ```CoreRuntime.Core60``` if there was a desire to be explicit.

### DIFF
--- a/tests/FSharpLint.Benchmarks/Program.fs
+++ b/tests/FSharpLint.Benchmarks/Program.fs
@@ -10,6 +10,6 @@ let main _ =
     BenchmarkRunner
         .Run<Benchmark>(
             DefaultConfig.Instance
-                .AddJob(Job.Default.WithRuntime(CoreRuntime.Core21))
+                .AddJob(Job.Default)
                 .AddDiagnoser(EtwProfiler())) |> ignore
     0


Numpsy:
Anyway, if i run it locally with this change it runs and I get results like
```
BenchmarkDotNet v0.13.11, Windows 10 (10.0.19045.3930/22H2/2022Update) (Hyper-V)
11th Gen Intel Core i7-1185G7 3.00GHz, 1 CPU, 4 logical and 2 physical cores
.NET SDK 6.0.418
  [Host]     : .NET 6.0.26 (6.0.2623.60508), X64 RyuJIT AVX2 DEBUG
  Job-HPIXNA : .NET 6.0.26 (6.0.2623.60508), X64 RyuJIT AVX2

Runtime=.NET 6.0

| Method         | Mean     | Error   | StdDev   | Median   |
|--------------- |---------:|--------:|---------:|---------:|
| LintParsedFile | 346.3 ms | 7.19 ms | 20.05 ms | 338.1 ms |
```

knocte:
Thanks!

PR #697, opened 2024-02-13
Fix AvoidSinglePipeOperator false negative in all inner expressions

Mersho:


knocte:
@webwarrior-ws please review

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -37,16 +37,53 @@ let runner (args: AstNodeRuleParams) =
         | _ ->
             Array.empty
 
+    let rec checkInnerExpressions (expr: SynExpr) =
+        match expr with
+        | SynExpr.App(_exprAtomicFlag, _isInfix, funcExpr, _argExpr, _range) ->
+            checkExpr funcExpr
+        | SynExpr.IfThenElse(_, thenExpr, elseExpr, _, _, _, _) ->
+            let checkThen = checkInnerExpressions thenExpr
```
webwarrior-ws, line 45: Since it's not a function, it's better to use noun as a name, not verb.

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -37,16 +37,53 @@ let runner (args: AstNodeRuleParams) =
...
+                |> Array.append (checkInnerExpressions elseExpression)
+            | None -> checkThen
+        | SynExpr.ArrayOrListComputed(_, expr, _) ->
+            checkInnerExpressions expr
+        | SynExpr.Sequential (_, _, firstExpression, secondExpression, _) ->
+            let result = checkInnerExpressions firstExpression
+            Array.append result (checkInnerExpressions secondExpression)
+        | SynExpr.Paren (innerExpression, _, _, _) ->
+            checkInnerExpressions innerExpression
+        | SynExpr.While (_, _, innerExpression, _) ->
```
webwarrior-ws, line 59: Also check condition expression, not only body

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -37,16 +37,53 @@ let runner (args: AstNodeRuleParams) =
...
+            checkInnerExpressions innerExpression
+        | SynExpr.For (_, _, _, _, _, _, _, innerExpression, _) ->
```
webwarrior-ws, line 61: same here

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -37,16 +37,53 @@ let runner (args: AstNodeRuleParams) =
...
+            checkInnerExpressions innerExpression
+        | SynExpr.ForEach (_, _, _, _, _, _, innerExpression, _) ->
```
webwarrior-ws, line 63: and here

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -37,16 +37,53 @@ let runner (args: AstNodeRuleParams) =
...
+            checkInnerExpressions innerExpression
+        | SynExpr.Match (_, _, clauses, _, _) ->
```
webwarrior-ws, line 65: and here - check also matched expression, not only clauses

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -37,16 +37,53 @@ let runner (args: AstNodeRuleParams) =
...
+            |> Array.append (checkInnerExpressions tryExpression)
+        | SynExpr.TryFinally (tryExpression, finallyExpr, _, _, _, _) ->
+            checkInnerExpressions finallyExpr
+            |> Array.append (checkInnerExpressions tryExpression)
+        | _ ->
+            Array.empty
+
     let error =
         match args.AstNode with
         | AstNode.Binding (SynBinding(_synAcc, _synBinding, _mustInline, _isMut, _synAttribs, _preXmlDoc, _synValData, _headPat, _synBindingRet, synExpr, _range, _debugPointAtBinding, _)) ->
```
webwarrior-ws, line 85: Is there a reason why only bindings are checked? What happens if pipeline operator is used in expression that is not a part of binding?
Like
```
module Foo

-1.0 |> abs
```

Maybe if you match `AstNode.Expression` instead, that big `checkInnerExpressions` function will be unnecessary?

knocte: sure, let's do this, but let's add a test first (and one that doesn't generate errors or warnings, because `abs` doesn't return unit, so it would generate a warning AFAIU)

knocte:
@webwarrior-ws  all good now?

src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -29,7 +35,14 @@ let runner (args: AstNodeRuleParams) =
                     | SynExpr.IfThenElse _ ->
                         Array.empty
                     | _ ->
-                        errors ident.idRange
+                        let isIdentPiped =
```
webwarrior-ws, line 38: `isIdentPied` is not a good name. It's not about identifier.

knocte:
@Mersho please re-run CI build of the commit titled "AvoidSinglePipeOperator: fix for all inner exprs"

Mersho:
> @Mersho please re-run CI build of the commit titled "AvoidSinglePipeOperator: fix for all inner exprs"

Done.

PR #699, opened 2024-02-15
Convert each FAKE target into its own .fsx script

Mersho:


knocte:
Please reopen PR when CI is green.

PR #703, opened 2024-03-10
update doc FL0071.md (cyclomaticComplexity)

Darkle:
Updated the rule settings json as it didn't inlcude the `cyclomaticComplexity` key.


knocte:
Thanks!

Issue #687, opened 2024-01-21
Execution failed for .net 6 project


plainionist:
I installed F# lint as dotnet tool as described in the documentation (version 0.24.0) and then tried to run it for a single fsproj on the command line like this:

dotnet fsharplint lint TheProject.fsproj

and i got:

Lint failed while analysing system\Bake\Bake\Bake.fsproj.
Failed with: Could not load file or assembly 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
Stack trace:    at Ionide.ProjInfo.ProjectLoader.loadProject(String path, BinaryLogGeneration binaryLogs, FSharpList`1 globalProperties)
   at Ionide.ProjInfo.ProjectLoader.getProjectInfo(String path, FSharpList`1 globalProperties, BinaryLogGeneration binaryLogs, FSharpList`1 customProperties) in /_//src/Ionide.ProjInfo/Library.fs:line 641
   at <StartupCode$Ionide-ProjInfo>.$Library.loadProject@906(WorkspaceLoader __, FSharpList`1 customProperties, BinaryLogGeneration binaryLogs, Dictionary`2 cache, String p) in /_//src/Ionide.ProjInfo/Library.fs:line 907
   at <StartupCode$Ionide-ProjInfo>.$Library.loadProjectList@934(WorkspaceLoader __, FSharpList`1 customProperties, BinaryLogGeneration binaryLogs, Dictionary`2 cache, FSharpList`1 projectList) in /_//src/Ionide.ProjInfo/Library.fs:line 944
   at Ionide.ProjInfo.WorkspaceLoader.Ionide.ProjInfo.IWorkspaceLoader.LoadProjects(FSharpList`1 projects, FSharpList`1 customProperties, BinaryLogGeneration binaryLogs) in /_//src/Ionide.ProjInfo/Library.fs:line 952
   at FSharpLint.Application.Lint.getProjectInfo(String projectFilePath, ToolsPath toolsPath) in /home/runner/work/FSharpLint/FSharpLint/src/FSharpLint.Core/Application/Lint.fs:line 301
   at FSharpLint.Application.Lint.lintProject$cont@396(OptionalLintParameters optionalParams, String projectFilePath, ToolsPath toolsPath, Unit unitVar) in /home/runner/work/FSharpLint/FSharpLint/src/FSharpLint.Core/Application/Lint.fs:line 440
[...]

PR #704, opened 2024-03-13
Try setting RollForward on FSharpLint.Console to latestMajor

Numpsy:
refs #687 , whereby it appears to not be able to process .NET 7 or 8 projects when running on .NET 6, and always using the latest version seems to help.

It looks like the current roll foward behaviour was added for #519 to fix issues where only a newer version than the tool was built with is available, but this change would make it just use the latest all the time (so for example, if you have 6 and 8 installed, then it will use 8 all the time).
I'm not sure if there is any potential to break anything that currently works though.


knocte:
Hey @Numpsy thanks for the PR!

> refs https://github.com/fsprojects/FSharpLint/issues/687 , whereby it appears to not be able to process .NET 7 or 8 projects when running on .NET 6, and always using the latest version seems to help.
> 
> It looks like the current roll foward behaviour was added for https://github.com/fsprojects/FSharpLint/issues/519 to fix issues where only a newer version than the tool was built with is available, but this change would make it just use the latest all the time (so for example, if you have 6 and 8 installed, then it will use 8 all the time).
> I'm not sure if there is any potential to break anything that currently works though.

Can you please include all this great info in the commit message please? commit messages can have body, not just a title.

Also, can you bring this to your PR: https://github.com/fsprojects/FSharpLint/issues/687#issuecomment-1949979932 This way we know for sure if the bug is fixed or not.

Numpsy:
I tried pulling in those other commits and the build failed, but in the logs I saw 
image

Which looks like it was downloading the artifact for the test 0.24.3 build, but then installing the 0.24.2 build from NuGet. I've tried to fix that by adding ```--prerelease``` to the install command in the test.

Numpsy:
Failure now looks like the ```testReleaseBinariesInWindows8``` step needs to do a paket restore before running the tool?

Numpsy:
Ok, so it now seems to be running the tool ok, but failing the build due to an error exit code, which is something i suppose:

image

knocte:
Good progress Richard! But why would it throw this error here and not throw it in our SelfCheck CI step?

.github/workflows/build+test+deploy.yml
```diff
@@ -89,6 +94,43 @@ jobs:
...
+    - name: Install FSharpLint from downloaded binaries
+      run: |
+        ls .\artifacts
+        dotnet tool install --global dotnet-fsharplint --prerelease --add-source .\artifacts
+    
+    - name: Setup Paket
+      run: |
+        dotnet tool install --global paket --version 7.2.1
+        paket restore
+
```
xperiandri, line 123: Probably we can remove this now?

xperiandri (changes requested):
Could you rebase and update to .NET 9?

Numpsy:
Hmm, if I try to run the unit tests on the lastest source with the preview .NET 10 SDK installed then I get a few test failures with messages like this:
```
FSharpLint.FunctionalTest test failed with 9 error(s) (63.9s)
    S:\GitHubForks\FSharpLint\tests\FSharpLint.FunctionalTest\TestApi.fs(87): error TESTERROR:
      Lint multi-targeted project (1s 242ms): Error Message:   MSBuildFailedToLoadProjectFile
        ("S:\GitHubForks\FSharpLint\tests\FSharpLint.FunctionalTest.TestedProject\FSharpLint.FunctionalTest.TestedProject.NetCore\FSharpLint.FunctionalTest.TestedP
      roject.NetCore.fsproj",
         InvalidProjectFileMessage
           "GenericError
        ("S:\GitHubForks\FSharpLint\tests\FSharpLint.FunctionalTest.TestedProject\FSharpLint.FunctionalTest.TestedProject.NetCore\FSharpLint.FunctionalTest.TestedP
      roject.NetCore.fsproj",
         "SDK Resolver Failure: "The SDK resolver "Microsoft.DotNet.MSBuildWorkloadSdkResolver" failed while attempting to resolve the SDK "Microsoft.NET.SDK.Workl
      oadAutoImportPropsLocator". Exception: "System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime, Version=10.0.0.0, Culture=neutral,
      PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
      File name: 'System.Runtime, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
         at Microsoft.NET.Sdk.WorkloadMSBuildSdkResolver.WorkloadSdkResolver.Resolve(SdkReference sdkReference, SdkResolverContext resolverContext, SdkResultFactor
      y factory)
         at Microsoft.Build.BackEnd.SdkResolution.SdkResolverService.TryResolveSdkUsingSpecifiedResolvers(IReadOnlyList`1 resolvers, Int32 submissionId, SdkReferen
      ce sdk, LoggingContext loggingContext, ElementLocation sdkReferenceLocation, String solutionPath, String projectPath, Boolean interactive, Boolean isRunningI
      nVisualStudio, SdkResult& sdkResult, IEnumerable`1& errors, IEnumerable`1& warnings)""  C:\Program Files\dotnet\sdk\10.0.100-preview.5.25277.114\Sdks\Microso
      ft.NET.Sdk\targets\Microsoft.NET.Sdk.ImportWorkloads.props")")
        Expected: True
        But was:  False
```

It seems ok if I set global.json to use the ,NET 9 SDK

knocte:
@Numpsy can you rename title of this PR to convey more clearly what it achieves? Seems the current title explains the **what** but not the **why** (it explains the solution, but not the problem).

Issue #488, opened 2021-06-29
Documentation says nothing about fsharplint.json file


not-rusty:
### Description

The docs say almost nothing about *which name does the configuration file have*, only [here](https://fsprojects.github.io/FSharpLint/how-tos/msbuild-task.html) implicitly. Also, it is also not obvious that in order to customize the linting I should copy the defaults listed [here](https://github.com/fsprojects/FSharpLint/blob/master/src/FSharpLint.Core/DefaultConfiguration.json), put them into a `fsharplint.json` file and edit them.

Coming from `eslint` I was expecting something like extending and overriding the defaults, like css.

Btw, thank you very much for this tool!

duckmatt:
Hi @not-rusty, thanks for raising this, that is some important info missing from the docs - added that in #489 

Regarding the overrides I agree that's how it should work, there was a regression when we moved from xml to json and it hasn't been addressed since, I'll try to take a look as part of this ticket

PR #711, opened 2024-06-15
Draft: Implement proper rule overriding to address issue #488

BennieCopeland:
I'm looking to add proper overriding back in. I expect that this will be a breaking change as the current functionality appears to allow running no rules by providing an empty JSON object. The new functionality will run all default enabled rules unless explicitly disabled in a custom config.

For the implementation, there are a few different ways I've considered that require discussion.

1. Keep the default rules in a JSON file, or rewrite them as F# code?
   - JSON File Pros
     - Can be used as an example for providing overrides
     - Single location for changing rules
   - JSON File Cons
     - Default rules are not co-located with the code implementing the rules
     - Code implementing the rules and the settings can become out of sync

   - F# Code Pros
     - Defaults can be co-located with the rules making them easier to keep updated with implementation changes
     - Out of sync defaults becomes a compile time issue
     - Splits the internal and external representations. Right now there there is a single Configuration type that includes deprecated settings. Splitting into Configuration and Override types can keep the internal code cleaner/more focused
   - F# Code Cons
     - Not as easy to show an example of a valid JSON config, but this could be mitigated by including an example config that is used by the FSharp.Data Json type provider. This will allow providing an example overrides file that also stays synchronized with the code base

   Whether the rules are kept in JSON or F#, adding/removing rules or changing defaults results in a new build and version bump.

2. Continue to use the same `Configuration` type, or rename `Configuration` to `Overrides` and create a new `Configuration` type without `optional` to reflect the default config.

### DIFF
--- a/src/FSharpLint.Core/Application/Configuration.fs
+++ b/src/FSharpLint.Core/Application/Configuration.fs
@@ -596,7 +596,7 @@ let loadConfig (configPath:string) =
 let defaultConfiguration =
     let assembly = typeof<Rules.Rule>.GetTypeInfo().Assembly
     let resourceName = Assembly.GetExecutingAssembly().GetManifestResourceNames()
-                       |> Seq.find (fun n -> n.EndsWith("fsharplint.json", System.StringComparison.Ordinal))
+                       |> Seq.find (fun n -> n.EndsWith(SettingsFileName, System.StringComparison.Ordinal))
     use stream = assembly.GetManifestResourceStream(resourceName)
     match stream with
     | null -> failwithf "Resource '%s' not found in assembly '%s'" resourceName (assembly.FullName)
--- a/src/FSharpLint.Core/Application/Lint.fs
+++ b/src/FSharpLint.Core/Application/Lint.fs
@@ -383,7 +383,7 @@ module Lint =
             | ex -> Error (string ex)
         | Default ->
             try
-                Configuration.loadConfig "./fsharplint.json"
+                Configuration.loadConfig $"./{Configuration.SettingsFileName}"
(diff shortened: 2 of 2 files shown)


BennieCopeland:
For the JSON config, if there are both deprecated configs and non-deprectated configs that do the same thing, which one take priority?

BennieCopeland:
Thinking about it, if the config is in F#, a new command line argument can be added that JSON serializes the type to the screen or a file for someone to start with.

BennieCopeland:
While comparing the `fsharplint.json` file to the `Configuration` type, I found some discrepancies.

Missing on the `Configuration` type
- unnestedFunctionNames
- nestedFunctionNames

Missing in the JSON file
- NonPublicValuesNames

BennieCopeland:
Pretty sure these lines are not correct for creating the `NonPublicValuesNames` rules: https://github.com/fsprojects/FSharpLint/blob/873d1452a0151f5541ff5470d7558cd0d6edd036/src/FSharpLint.Core/Application/Configuration.fs#L711-L712

BennieCopeland:
The following rules are part of the `allRules` array when flattening the config:

- UnnestedFunctionNames
- NestedFunctionNames
- SuggestUseAutoProperty

knocte:
Hi Bennie, thanks for your work so far! Sorry for the delay replying, life is kind of hectic nowadays for me.

> The following rules are NOT part of the allRules array when flattening the config and thus never being ran:

Do you mind opening a separate PR to fix this please?

BennieCopeland:
Sure, new PR for that change here: #713.

PR #715, opened 2024-09-26
Update FAKE

Thorium:
This update FAKE and PAKET build tools to support more recent software configurations (like building with machine having .NET 9 RC installed).


Thorium:
This was just smaller pre-commit needed for Net8-build. Either merge this first or not at all.

Numpsy:
I think everything in this one has been merged or superceeded now

Issue #618, opened 2023-11-23
Failed with: "The exception has been reported."


Thorium:
Trying the tool with one sample:

```
git clone https://github.com/simontreanor/FunStripe.git
cd FunStripe
cd src
dotnet fsharplint lint FunStripe.fsproj
```

Expected: Run and produce a lot of issues.

Actual:
```
Lint failed while analysing FunStripe.fsproj.
Failed with: The exception has been reported. This internal exception should now be caught at an error recovery point on the stack. Original message: One or more errors seen during provided type setup)
Stack trace:    at FSharp.Compiler.ErrorLogger.ErrorLoggerExtensions.ErrorLogger.Error[b](ErrorLogger x, Exception exn) in D:\workspace\_work\1\s\src\fsharp\ErrorLogger.fs:line 390
   at FSharp.Compiler.CompilerImports.TcImports.ImportTypeProviderExtensions(CompilationThreadToken ctok, TcConfig tcConfig, String fileNameOfRuntimeAssembly, ILScopeRef ilScopeRefOfRuntimeAssembly, FSharpList`1 runtimeAssemblyAttributes, Entity entityToInjectInto, FSharpEvent`1 invalidateCcu, Range m) in D:\workspace\_work\1\s\src\fsharp\CompilerImports.fs:line 1395
   at FSharp.Compiler.CompilerImports.ccuRawDataAndInfos@1566-1.Invoke(Unit unitVar0) in D:\workspace\_work\1\s\src\fsharp\CompilerImports.fs:line 1570
   at FSharp.Compiler.CompilerImports.PrepareToImportReferencedFSharpAssembly@1581.Invoke(Unit unitVar0) in D:\workspace\_work\1\s\src\fsharp\CompilerImports.fs:line 1590
   at Microsoft.FSharp.Primitives.Basics.List.collectToFreshConsTail[T,TResult](FSharpFunc`2 f, FSharpList`1 list, FSharpList`1 cons) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\local.fs:line 439
   at Microsoft.FSharp.Primitives.Basics.List.collect[T,TResult](FSharpFunc`2 f, FSharpList`1 list) in D:\workspace\_work\1\s\src\fsharp\FSharp.Core\local.fs:line 447
[...]

Issue #566, opened 2022-08-24
`dotnet fsharplint lint ...` fails to run on a machine with only .net X.0 installed, X > 5


SteveGilham:
### Description

`dotnet fsharplint lint ...` fails to run on a machine with only .net 6.0 installed

### Repro steps

1) dotnet tool install dotnet-fsharplint
2) Remove all .net SDKs prior to current (6.0.400 at time of writing)
3) Invoke `dotnet fsharplint lint` on any .fsproj file

### Expected behavior

Tool runs to analyze project

### Actual behavior

```
Unhandled exception. System.TypeInitializationException: The type initializer for 'FSharpLint.Console.Program' threw an exception.
 ---> System.TypeInitializationException: The type initializer for '<StartupCode$dotnet-fsharplint>.$FSharpLint.Console.Program' threw an exception.
 ---> System.InvalidOperationException: No instances of MSBuild could be detected.
Try calling RegisterInstance or RegisterMSBuildPath to manually register one.
   at Microsoft.Build.Locator.MSBuildLocator.RegisterDefaults()
   at Ionide.ProjInfo.Init.init() in D:\a\proj-info\proj-info\src\Ionide.ProjInfo\Library.fs:line 17
   at <StartupCode$dotnet-fsharplint>.$FSharpLint.Console.Program..cctor() in D:\a\FSharpLint\FSharpLint\src\FSharpLint.Console\Program.fs:line 134
   --- End of inner exception stack trace ---
   at FSharpLint.Console.Program..cctor()
   --- End of inner exception stack trace ---
   at FSharpLint.Console.Program.main(String[] argv) in D:\a\FSharpLint\FSharpLint\src\FSharpLint.Console\Program.fs:line 138
```

### Known workarounds

Install out-of-support .net 5.0.408 SDK

### Related information

* Operating system : Windows
* Branch : nuget package dotnet-fsharplint version 0.21.2
* .NET Runtime, CoreCLR or Mono Version : .net SDK 6.0.400

SteveGilham:
Seems to be a duplicate of #536

SteveGilham:
Probably the simplest fix would be to detect (or specify via command line) the SDK in use & feed that into a manual resolution process if the default doesn't work.

baronfel:
It would probably be easiest to update the proj-info dependencies, which I believe would also require updating FSharp.Compiler.Service as well. Newer proj-info dependency versions bring better resolution logic here, and better error messages.

[...]

PR #716, opened 2024-09-26
Build with .NET 8, support .NET 8 projects

Thorium:
Build with .NET 8.
Support analysing .NET 8 project (like itself now).
Updated build-tools and Security vulnerability fix in Asn1 dependency

Ping #687 #618 #566 #604 et al.


C0DK:
If this actually makes NET 8 work, that'd be great to merge :)

Numpsy:
> Security vulnerability fix in Asn1 dependency

fwiw it might be possible to remove some references to the System.Formats.Asn1 NuGet package altogether by updating System.Security.Cryptography.Pkcs to 8.0.1 (The .NET 8 TFM in 8.0.0 depends on the System.Formats.Asn1 v8 NuGet package, but in 8.0.1 that reference was removed)

PR #722, opened 2025-06-29
Migrated to .NET `9.0.201` and F# compiler `43.9.201`

xperiandri:
* Removed `Paket`
* Migrated to .NET `9.0.201` and F# compiler `43.9.201`
* Implemented missing AST constructs handling
* Fixed tests


Directory.Packages.props
```diff
@@ -0,0 +1,30 @@
...
+    <PackageVersion Include="BenchmarkDotNet" Version="0.15.2" />
+    <PackageVersion Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.15.2" />
+    <PackageVersion Include="FParsec" Version="1.1.1" />
+    <PackageVersion Include="FSharp.Compiler.Service" Version="43.9.201" />
+    <PackageVersion Include="FSharp.Control.Reactive" Version="6.1.2" />
+    <PackageVersion Include="FSharp.Core" Version="9.0.201" />
+    <PackageVersion Include="Ionide.ProjInfo.ProjectSystem" Version="0.70.2" />
+    <PackageVersion Include="Ionide.ProjInfo.FCS" Version="0.70.2" />
+    <PackageVersion Include="Ionide.ProjInfo" Version="0.70.2" />
+    <PackageVersion Include="Ionide.ProjInfo.Sln" Version="0.68.0" />
```
Numpsy, line 17: fwiw, ProjInfo 0.70 dropped Ionide.ProjInfo.Sln in favor of using the Microsoft ```Microsoft.VisualStudio.SolutionPersistence``` instead, so I don't know if this is used directly?
Querying because one of the reasons for the change was to get support for .slnx solutions, which could be useful here?

xperiandri: Great suggestion. I've implemented that. It will come in another PR

PR #723, opened 2025-07-01
Added SLNX and SLNF formats support and migrated to SLNX solution

xperiandri:


xperiandri:
Great idea!

xperiandri:
@Numpsy implemented `slnf` support

PR #724, opened 2025-07-01
Use string interpolation instead of `sprintf` in tests

xperiandri:


docs/generators/lunr.fsx
```diff
@@ -42,8 +42,8 @@ let generate (ctx : SiteContents) (projectRoot: string) (page: string) =
 
           let gen =
               let ctn =
-                  sprintf "%s \n %s" generatorOutput.AssemblyGroup.Name (generatorOutput.AssemblyGroup.Namespaces |> Seq.map (fun n -> n.Name) |> String.concat " ")
-              {uri = (rootUrl + sprintf "/reference/%s/index.html" n.Label ); title = sprintf "%s - API Reference" n.Label; content = ctn }
+                  $"%s{generatorOutput.AssemblyGroup.Name} \n %s{(generatorOutput.AssemblyGroup.Namespaces |> Seq.map (fun n -> n.Name) |> String.concat " ")}"
```
Numpsy, line 45: The compiler doesn't seem to like the nested quotes in the ```String.concat " "``` here? 

image

src/FSharpLint.Console/Output.fs
```diff
@@ -34,29 +34,30 @@ type StandardOutput () =
...
         member _.WriteError (error:string) =  writeLine error ConsoleColor.Red Console.Error
 
 type MSBuildOutput () =
+
     interface IOutput with
+
         member _.WriteInfo (info:string) = Console.Out.WriteLine info
         member _.WriteWarning (warning:Suggestion.LintWarning) =
-            sprintf "%s(%d,%d,%d,%d):FSharpLint warning %s: %s"
+            fprintf Console.Out "%s(%d,%d,%d,%d):FSharpLint warning %s: %s"
```
Numpsy, line 53: @xperiandri Should this be ```fprintfn``` as it used to do Out.WriteLine?

(ref https://github.com/fsprojects/FSharpLint/issues/740#issuecomment-3062105171)

PR #727, opened 2025-07-01
Updated `.gitignore` and added `.editorconfig` with Fantomas settings

xperiandri:
* Use Fantomas and `.editorconfig` (my setup, I have not tried to align with the repo style)
* Use targets to automatically restore tools and automatically format the whole repo

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,6 +15,13 @@
         "fornax"
       ],
       "rollForward": false
+    },
+    "fantomas": {
+      "version": "7.0.2",
+      "commands": [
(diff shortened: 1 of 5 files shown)


knocte:
> If you prefer going this way instead of forcing the whole repo formatting.

I actually prefer whole repo formatting, however, I was holding off doing fantomas myself because I knew that doing it right now would make a bunch of PRs have conflicts; and not just straightforward conflicts, but conflicts very annoying to resolve. Therefore I would be inclined to work on this only after we have merged a bunch of PRs that are pending (yes the DRAFT ones also count, we plan to revisit those soon).

PR #731, opened 2025-07-03
update the Github CI to use .NET SDK 9.0.201

Numpsy:
The CI in my fork is still trying to run , but is failing because it installs the .NET 6 SDK which can't build the .NET 9 libs, so - update it

refs https://github.com/Numpsy/FSharpLint/actions/runs/16049171264/job/45287501138#step:5:9

### DIFF
--- a/.github/workflows/build+test+deploy.yml
+++ b/.github/workflows/build+test+deploy.yml
@@ -15,7 +15,7 @@ on:
       - '**'
 
 env:
-  DOTNET_VERSION: 6.0.417
+  DOTNET_VERSION: 9.0.201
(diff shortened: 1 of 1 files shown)


Numpsy:
Ye, the FAKE build script is still trying to use Paket, so it needs something like https://github.com/fsprojects/FSharpLint/pull/730 to change it to plain NuGet.

(That does then fail doing the 'SelfCheck' step, but I suppose that's a different issue - https://github.com/Numpsy/FSharpLint/actions/runs/16071143791/job/45355872393#step:8:40)

PR #732, opened 2025-07-04
Fixed rules violation in the linter code

xperiandri:
* disable `FL0046` for `Identifiers.fs`
* make constants instead of let bindings in `SourceLength.fs`

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -64,7 +64,7 @@ let isNullOrWhiteSpace = System.String.IsNullOrWhiteSpace
 let exec cmd args dir =
     let proc =
         CreateProcess.fromRawCommandLine cmd args
-        |> CreateProcess.ensureExitCodeWithMessage (sprintf "Error while running '%s' with args: %s" cmd args)
+        |> CreateProcess.ensureExitCodeWithMessage $"Error while running '%s{cmd}' with args: %s{args}"
     (if isNullOrWhiteSpace dir then proc
     else proc |> CreateProcess.withWorkingDirectory dir)
     |> Proc.run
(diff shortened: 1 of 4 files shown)


xperiandri:
@Numpsy could you review?

Numpsy:
I've been having a go at getting the Fornax doc generation working over in https://github.com/Numpsy/FSharpLint/commit/686d1399c3c62807ba3ea622e94e928ab43907af, not quite working yet though

knocte:
> disable FL0046 for Identifiers.fs

You explained the what but not the why. What was the motivation for doing this?

xperiandri:
Well, someone decided to name the identifiers in pascal case. Which is logical.
But F# does not allow compiling an interpolated string into a literal.
Probably I should have mentioned that

knocte:
> Probably I should have mentioned that

Or file that as a bug (with a testcase) so that we can take a closer look? I'm still not sure I understand the issue.

knocte:
> But F# does not allow compiling an interpolated string into a literal.

But what was the motivation for having constants (literals) in the first place?

xperiandri:
I do not know. Someone did that before me and you in this project

knocte:
I'm not referring to previous code, I'm referring to what you have put in this PR description: "make constants instead of let bindings in SourceLength.fs", what is the motivation of making them constants instead of let bindings?

xperiandri:
Because constants are Pascal case and let bindings are Camel case. But in the code let bindings are Camel case

Numpsy:
The previously seen issue might have been a bug that was fixed by https://github.com/fsprojects/FSharpLint/pull/758?

PR #733, opened 2025-07-04
Try to get the docs building with NuGet instead of Paket

Numpsy:
An attempt at getting the Fornax docs build to work without Paket (it expects various dlls to be in certain places, where they used to be put by Paket)

Borrows the approach from https://github.com/fsprojects/FSharpLint/pull/680 of using PackageDownload to get the packages, but isn't quite working in the CI yet with a failure i'm not clear on (I haven't used Fornax before)

### DIFF
--- a/docs/loaders/apirefloader.fsx
+++ b/docs/loaders/apirefloader.fsx
@@ -1,5 +1,5 @@
 #r "../_lib/Fornax.Core.dll"
-#r "../../packages/docs/FSharp.Formatting/lib/netstandard2.0/FSharp.MetadataFormat.dll"
+#r "../../packages/docs/fsharp.formatting/4.1.0/lib/netstandard2.0/FSharp.MetadataFormat.dll"
 
 open System
 open System.IO
@@ -38,7 +38,7 @@ let loader (projectRoot: string) (siteContet: SiteContents) =
           Path.Combine (projectRoot, "..", "build")
         ]
       for (label, dll) in dlls do
-        let output = MetadataFormat.Generate(dll, markDownComments = true, publicOnly = true, libDirs = libs)
+        let output = MetadataFormat.Generate(dllFile = dll, markDownComments = true, publicOnly = true, libDirs = libs, parameters = [("project-name", label)])
 
         let allModules =
             output.AssemblyGroup.Namespaces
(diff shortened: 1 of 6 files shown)


docs/loaders/apirefloader.fsx
```diff
@@ -38,7 +38,7 @@ let loader (projectRoot: string) (siteContet: SiteContents) =
           Path.Combine (projectRoot, "..", "build")
         ]
       for (label, dll) in dlls do
-        let output = MetadataFormat.Generate(dll, markDownComments = true, publicOnly = true, libDirs = libs)
+        let output = MetadataFormat.Generate(dllFile = dll, markDownComments = true, publicOnly = true, libDirs = libs, parameters = [("project-name", label)])
```
Numpsy, line 41: This is a go at fixing the 'no project name specified' error at https://github.com/Numpsy/FSharpLint/actions/runs/16079166194/job/45380884425#step:7:58

Numpsy:
The CI needs #732 before it gets far enough to test this - this works locally for me on Windows at least

Numpsy:
Will conflict with https://github.com/fsprojects/FSharpLint/pull/723

Numpsy:
The 'deplyReleaseBinaries' CI step seems to be running the unit tests 3 times - is that needed?

xperiandri:
I don't see it running 3 times in YAML definition. Could you show me where you have seen that?

Numpsy:
The build logs for the 'deplyReleseBinaries' step in https://github.com/fsprojects/FSharpLint/actions/runs/16099807422/job/45427611835 the logs make it look like the 'Build', 'Pack'  and 'Upload binaries to 'NuGet' steps are each running the unit tests

xperiandri:
Pretty much possible

xperiandri:
Won't it build with a higher version of SDK? Or 3.1+9?

Numpsy:
> Won't it build with a higher version of SDK? Or 3.1+9?

I was thinking that there could be issues due to the versions of Fornax and FSharp.Formatting that it uses having a quite old version of FSharp.Compiler.Service built in to it, which can't handle the newer language syntax (just an assumption though).

The other PR to use the newer Fornax seems a better idea ging forwards though (don't really want to be depending on .NET Core 3.1 tool at this point)

knocte:
> Hey @Numpsy sorry, seems I'm out of the loop but if Paket was removed...

Ah nevermind, I had only glimpsed only at the title and missed the word "docs". Ok, so this is also about Fornax.

PR #734, opened 2025-07-06
Update the build instructions to use `dotnet fsi` instead of `fake-cli`

Numpsy:
To match the updated CI.

Maybe completing https://github.com/fsprojects/FSharpLint/pull/660 would be usful, so you don't need to crea about FAKE or whatever if you just want to build the docs?


xperiandri (approved):

knocte:
> Maybe completing https://github.com/fsprojects/FSharpLint/pull/660 would be usful, so you don't need to crea about FAKE or whatever if you just want to build the docs?

Good point, I'm gonna update that PR.

PR #737, opened 2025-07-07
build.fsx: only push Nuget packages from master branch

webwarrior-ws:
To avoid running this step in situations like opening PR from main repository.


webwarrior-ws:
@knocte moved check for "master" branch after tag check

PR #742, opened 2025-07-10
Fake build project on .NET 9

xperiandri:
* Build project is on .NET 9
* `launchSettings.json` with multiple profiles to run the build project
* Target to execute locally and issue a release: update changelog, commit, set tag
* Publish successful merges to `master` as prerelease GitHub packages
* Embed all sources into GitHub packages for easy local debugging

### DIFF
--- a/build/build.fs
+++ b/build/build.fs
@@ -639,7 +639,7 @@ let selfCheck _ =
     DotNet.exec
         (fun opts -> { opts with WorkingDirectory = consoleProj })
         "run"
-        $"lint %s{sol}"
+        $"--framework net9.0 lint %s{sol}"
     |> failOnBadExitAndPrint
 
 let initTargets (ctx : Context.FakeExecutionContext) =


knocte:
> Fake build project on .NET 9

What does the title of this PR really mean? If with "Fake" you mean the build system FAKE, maybe rather put it in uppercase letters? Not that I'm trying to be a nitpicker but I don't understand what "FAKE build project on .NET9" really means if master branch is already targetting .NET9. What does this PR really achieve? Can you reword the title of PR to convey that message more clearly please?

I'll continue my review now, please don't merge yet until you got a green light from me because I still this is still reverting some improvements that we have made in the past.

.github/workflows/build+test+deploy.yml
```diff
@@ -55,45 +55,39 @@ jobs:
...
+        NUGET_TOKEN: ${{ secrets.NUGET_TOKEN }}
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      run: dotnet fsi build.fsx -t Push
-    - name: Create Release (if tag)
-      if: startsWith(github.ref, 'refs/tags/')
-      id: create_release
-      uses: actions/create-release@latest
+        FAKE_DETAILED_ERRORS: true
+
+      run: ./build.ps1 Publish
```
knocte, line 81: @xperiandri why change the publish job to use Windows instead of Linux? what's the motivation?

.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,34 @@
+name: "Copilot Setup Steps"
```
knocte, line 1: @xperiandri looks like this file is unrelated to this PR?

knocte:
> I'll continue my review now

I'm having some connection issues atm, will continue tomorrow.

.github/workflows/build+test+deploy.yml
```diff
@@ -55,45 +55,39 @@ jobs:
...
+
+    - name: PublishToGitHub (if master branch)
+      if: github.ref == 'refs/heads/master'
```
knocte, line 84: @xperiandri this logic was moved in the past from GithubActions to F#, let's keep it in F# please

knocte:
> @xperiandri this logic was moved in the past from GithubActions to F#, let's keep it in F# please

The reason is one that I kind of hinted in the previous PR already: if you make the Publish CI job only be run in master branch, then changing the job might lead to breaking it without noticing it in the PR status. However, doing the check in F# instead of with GitHubActions' `if` statement allows the job to run always, except for the last step (the nuget push). This way, the `nuget pack` part gets exercised everytime (and it's not rate to break nuget pack operations by mistake if you, for example, change some properties of the .fsproj files).

xperiandri:
I don't understand how your last comment is related to this PR.
> The reason is one that I kind of hinted in the previous PR already: if you make the Publish CI job only be run in master branch, then changing the job might lead to breaking it without noticing it in the PR status. However, doing the check in F# instead of with GitHubActions' if statement allows the job to run always, except for the last step (the nuget push). This way, the nuget pack part gets exercised everytime (and it's not rare to break nuget pack operations by mistake if you, for example, change some properties of the .fsproj files).

The pipeline contains `./build.ps1 DotnetPack` which is not constrained by any condition.

For me, it looks strange to use Fake but call targets separately in the pipeline. If we use Fake the pipeline should consist of.NET install and calling Fake. 
The current pipeline does
``` YAML
      run: dotnet restore
    - name: Build
      run: dotnet fsi build.fsx -t Build
    - name: Run tests
      run: dotnet fsi build.fsx -t Test
    - name: Run FSharpLint on itself
      run: dotnet fsi build.fsx -t SelfCheck
```

but inside we have
``` F#
"Clean"
  ==> "Build"
  ==> "Test"
  ==> "Default"
```

which means that we build twice, and here

``` F#
"Default"
  ==> "Pack"
  ==> "Push"
  ==> "Release"
```
 with
 
 ``` YAML
     - name: Build
      run: dotnet fake build
    - name: Pack
      run: dotnet fake build -t Pack
```

we build twice again. As @Numpsy told


And what exists now is
``` YAML
    - name: Upload binaries to nuget (if nugetKey present)
      env:
        nuget-key: ${{ secrets.NUGET_KEY }}
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      run: dotnet fake build -t Push
    - name: Create Release (if tag)
      if: startsWith(github.ref, 'refs/tags/')
      id: create_release
      uses: actions/create-release@latest
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
      with:
        tag_name: ${{ github.ref }}
        release_name: ${{ github.ref }}
        body: ${{ steps.changelog_reader.outputs.log_entry }}
        draft: false
        prerelease: false
    - name: Upload binaries to release (if tag)
      if: startsWith(github.ref, 'refs/tags/')
      uses: svenstaro/upload-release-action@v1-release
      with:
        repo_token: ${{ secrets.GITHUB_TOKEN }}
        file: out/*.nupkg
        tag: ${{ github.ref }}
        overwrite: true
        file_glob: true
```

and this logic is extremely complicated and impossible to understand without GPT explanation

``` F#
    let key = getBuildParam "nuget-key"
    match getBuildParam "GITHUB_EVENT_NAME" with
    | None ->
        match key with
        | None ->
            let key = UserInput.getUserPassword "NuGet Key: "
            push key
        | Some key ->
            push key

    | Some "push" ->
        match key with
        | None ->
            Console.WriteLine "No nuget-key env var found, skipping..."
        | Some key ->
            if isTag then
                push key
            else
                match getBuildParam "GITHUB_SHA" with
                | None ->
                    failwith "GITHUB_SHA should have been populated"
                | Some commitHash ->
                    let gitArgs = sprintf "describe --exact-match --tags %s" commitHash
                    let proc =
                        CreateProcess.fromRawCommandLine "git" gitArgs
                        |> Proc.run
                    if proc.ExitCode <> 0 then
                        // commit is not a tag, so go ahead pushing a prerelease
                        push key
                    else
                        Console.WriteLine "Commit mapped to a tag, skipping pushing prerelease..."
    | _ ->
        Console.WriteLine "Github event name not 'push', skipping..."
```

I think it is a crazy mess.
________________________________

I have clean working code and pipelines from https://github.com/TheAngryByrd/MiniScaffold implemented in 3 my repos and multiple others. Let it have a few adjustments that you want.


1. CI pipeline for everything does Pack
2. Publish to GitHub packages from `master`
3. Publish to NuGet from `v*` tag

All pipelines look like
``` YAML
    - uses: actions/checkout@v4
    - name: Setup .NET
      uses: actions/setup-dotnet@v4
      with:
        global-json-file: global.json
    - name: Restore tools
      run: dotnet tool restore
    - name: Build
      shell: pwsh
      run: ./build.ps1 <call target>
```

Clean, easy to read and maintain. One pipeline has only a pipeline-level trigger condition and calls Fake.

knocte:
> we build twice again

First let me ask you this basic question: if you clone the project, and you try to run a target that runs tests, shouldn't the build system know that the project hasn't been built yet and then launch the build before trying to run the tests? I hope yes, then I would agree.

Now, if you agree with the above, it's only normal to think that other targets should behave similarly, e.g. if you instruct the build system to "Pack", then it should build first if there was no build before, right?

Alright, so now that we agree on basic functionality of a build system, let's talk performance! If you instruct the build system to "Build" first, and after that, you instruct it to "Pack", should it build twice? Well, if you're using a good build system, it shouldn't. And this is not a conversation about CI, this is just talking build systems. So let's see what happens if you use Make: what you normally do in this case is create a target with the same name as the binary file that gets generated by the build; after that you make that target depend on all the source files that are needed to generate that binary; and then you create a target called "Build" and make it depend on the source files and on the binary file. This way, if you instruct Make to build twice (e.g. running "make" twice in your terminal) it will actually only run the build once! Because it detects that the source files didn't change after the last binary was generated. After this, you could create a target for packing, and so on.

Now, does FAKE work this way to avoid building twice? I'm not sure! I told you I'm not a fan of FAKE (and not a fan of MSBuild either, truth to be told) so I didn't learn it enough to know if this is possible. But if this is possible, then we should be able to make use of the FAKE feature that allows this, without having to change our CI steps (remember, we have 2 steps: one for building, another for packing, so that when it fails, we know which part failed! if we didn't care about this information, then, for sure, we would only need a CI step that just packs).

I hope you understood where I'm coming from here.

> I think it is a crazy mess.

You're welcome to refactor that code if you think you can make it more readable. But to me, you're just overhauling a bunch of things here which are regressing in terms of what the build system can do right now in the master branch, for example: if we merged this PR as it is, and after that we created a new tag (a new version), would your new build scripts be able to figure out that the commit is a tag and, therefore, not push a pre-release version to nuget.org for that commit?

xperiandri:
I have no device without .NET 9 SDK installed to verify, but I suppose it fixes that

knocte:
> I have no device without .NET 9 SDK installed to verify, but I suppose it fixes that

Did you mean to write this comment in PR 748? Because I assume that with "without .NET 9" you mean "with only .net8", something which doesn't seem related to this PR.

xperiandri:
Sorry, wrong PR

Issue #123, opened 2015-09-16
Error on F# project mono / osx


7sharp9:
```
/Users/dave/code/xamarin/fsharpdesigner/src/packages/FSharpLint.0.2.6/build/FSharpLint.targets: Warning: Lint failed while analysing /Users/dave/code/xamarin/fsharpdesigner/src/Xamarin.Providers/Xamarin.Providers.fsproj.
Failed with: Unknown tools version: '12.0' . Known versions: '2.0' '3.0' '3.5' '4.0'
Stack trace:    at Microsoft.Build.BuildEngine.Engine.set_DefaultToolsVersion (System.String value) in <filename unknown>:line 0
   at Microsoft.FSharp.Compiler.SourceCodeServices.FSharpProjectFileInfo.CrackProjectUsingOldBuildAPI (System.String fsprojFile) in <filename unknown>:line 0
   at Microsoft.FSharp.Compiler.SourceCodeServices.FSharpProjectFileInfo..ctor (System.String fsprojFileName, Microsoft.FSharp.Core.FSharpOption`1 properties, Microsoft.FSharp.Core.FSharpOption`1 enableLogging) in <filename unknown>:line 0 (Xamarin.Providers)
```
Looks to me like an assembly redirect issue on msbuild resolution while creating the project information in FCS.

duckmatt:
Thanks, confusingly the acceptance test is passing on travis, does it run as expected when msbuild is run via the command line?

duckmatt:
`FSharp.Commpiler.Services` handles this now, so I believe it will have been fixed in https://github.com/fsprojects/FSharpLint/pull/153 which will be included in the next release

PR #743, opened 2025-07-10
Exit after printing out version information

Numpsy:
Rather than then trying to call GetSubCommand() and breaking when there isn't one.

refs https://github.com/fsprojects/FSharpLint/issues/741

Just a thought at a simple change to exit after printing the version information, rather than making a call to GetSubCommand() that can fail if there no other commands specified.

If it needs to support printing the version and then doing a lint, maybe it needs to use TryGetSubCommand() instead.

### DIFF
--- a/src/FSharpLint.Console/Program.fs
+++ b/src/FSharpLint.Console/Program.fs
@@ -85,8 +85,7 @@ let private start (arguments:ParseResults<ToolArgs>) (toolsPath:Ionide.ProjInfo.
             Assembly.GetExecutingAssembly().GetCustomAttributes false
             |> Seq.pick (function | :? AssemblyInformationalVersionAttribute as aiva -> Some aiva.InformationalVersion | _ -> None)
         $"Current version: {version}" |> output.WriteInfo
-        Environment.Exit(0)
-    
+        Environment.Exit 0
     let handleError (str:string) =
         output.WriteError str
         exitCode <- -1
--- a/src/FSharpLint.Console/Program.fs
+++ b/src/FSharpLint.Console/Program.fs
@@ -85,8 +85,8 @@ let private start (arguments:ParseResults<ToolArgs>) (toolsPath:Ionide.ProjInfo.
             Assembly.GetExecutingAssembly().GetCustomAttributes false
             |> Seq.pick (function | :? AssemblyInformationalVersionAttribute as aiva -> Some aiva.InformationalVersion | _ -> None)
         $"Current version: {version}" |> output.WriteInfo
-        ()
-
+        Environment.Exit(0)
+    
     let handleError (str:string) =
         output.WriteError str
         exitCode <- -1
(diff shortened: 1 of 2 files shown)


xperiandri:
I agree

xperiandri:
@webwarrior-ws @Numpsy do you have any remarks?

xperiandri:
@Numpsy, what do you think about the suggestion?

xperiandri (approved):

knocte:
> refs https://github.com/fsprojects/FSharpLint/issues/741

Not sure the text "refs #123" makes the closing of #123 be automatic, maybe it needs to be "Fixes", not "refs"?
BTW, please also use this in commit msg, not just in PR description.

src/FSharpLint.Console/Program.fs
```diff
@@ -85,8 +85,8 @@ let private start (arguments:ParseResults<ToolArgs>) (toolsPath:Ionide.ProjInfo.
             Assembly.GetExecutingAssembly().GetCustomAttributes false
             |> Seq.pick (function | :? AssemblyInformationalVersionAttribute as aiva -> Some aiva.InformationalVersion | _ -> None)
         $"Current version: {version}" |> output.WriteInfo
-        ()
-
+        Environment.Exit(0)
```
knocte, line 88: @Numpsy nit: no need for parentheses

src/FSharpLint.Console/Program.fs
```diff
@@ -85,8 +85,8 @@ let private start (arguments:ParseResults<ToolArgs>) (toolsPath:Ionide.ProjInfo.
...
+
```
knocte, line 89: @Numpsy nit: don't add trailing whitespace please

knocte (changes requested):

src/FSharpLint.Console/Program.fs
xperiandri:
```suggestion
        Environment.Exit 0
```

src/FSharpLint.Console/Program.fs
xperiandri:
```suggestion

```

src/FSharpLint.Console/Program.fs
```diff
@@ -85,8 +85,7 @@ let private start (arguments:ParseResults<ToolArgs>) (toolsPath:Ionide.ProjInfo.
             Assembly.GetExecutingAssembly().GetCustomAttributes false
             |> Seq.pick (function | :? AssemblyInformationalVersionAttribute as aiva -> Some aiva.InformationalVersion | _ -> None)
         $"Current version: {version}" |> output.WriteInfo
-        ()
-
+        Environment.Exit 0
```
xperiandri, line 88:
```suggestion
        Environment.Exit 0

```

xperiandri (approved):

PR #759, opened 2025-08-06
NestedFunctionNames,UnnestedFunctionNames: fix rules

webwarrior-ws:
Fixed rules to only take into account function and method names.

Added test cases make sure the rule doesn't fire on bindings that are not functions.


knocte:
Let's also add a test (for each rule) that has a function with no parameters (so, with the unit `()`) param, where the rule should give a positive.

webwarrior-ws:
> Let's also add a test (for each rule) that has a function with no parameters (so, with the unit `()`) param, where the rule should give a positive.

But we already have such tests.

knocte:
I just checked tests/FSharpLint.Core.Tests/Rules/Conventions/Naming/NestedFunctionNames.fs and all testcases with nested funcs have one param.

PR #761, opened 2025-08-11
Update the 'Running Lint From An Application' documentation to say in…

Numpsy:
…stall FSharpLint.Core rather than FSharp.Core

As it stands, the text says 'FSharp.Core' but the link goes to FSharpLint.Core, and It looks like they should both say FSharpLint.Core?


Numpsy:
refs 
image


from https://fsprojects.github.io/FSharpLint/

knocte:
Thanks @Numpsy

PR #765, opened 2025-08-17
Core(HintParser): Change charListToString to use String.Concat

Numpsy:
Just a thought I had after running a run of FSharpLint over one of my projects thought the Visual Studio memory profiler. It's quite trivial in the overall schema of things, but in case anyone is interested.

I noticed that ```charListToString``` was creating quite a large number of strings:

Screenshot 2025-08-10 224138

And I wondered if ```String.Create``` could be used to reduce the number of temporary strings. Doing so got

Screenshot 2025-08-11 005042

The actual amounts of memory here are rather small compared to total memory because the strings are tine - 

Before
```
| Method      | Mean    | Error    | StdDev   | Gen0        | Gen1       | Gen2      | Allocated  |
|------------ |--------:|---------:|---------:|------------:|-----------:|----------:|-----------:|
| LintProject | 3.983 s | 0.1033 s | 0.2964 s | 167000.0000 | 22000.0000 | 1000.0000 | 1008.65 MB |
```

After
```
| Method      | Mean    | Error    | StdDev   | Gen0        | Gen1       | Gen2      | Allocated  |
|------------ |--------:|---------:|---------:|------------:|-----------:|----------:|-----------:|
| LintProject | 3.916 s | 0.1088 s | 0.3088 s | 167000.0000 | 22000.0000 | 1000.0000 | 1007.31 MB |
```

But the change is also small so maybe it's still useful.

### DIFF
--- a/src/FSharpLint.Core/Framework/HintParser.fs
+++ b/src/FSharpLint.Core/Framework/HintParser.fs
@@ -7,8 +7,8 @@ open HintParserTypes
 
 module HintParser =
 
-    let charListToString charList =
-        Seq.fold (fun concatenatedString charElement -> concatenatedString + charElement.ToString()) String.Empty charList
+    let charListToString (charList: char seq) =
+        String.Concat charList
 
     let pischar chars : Parser<char, 'CharParser> =
         satisfy (fun character -> List.exists ((=) character) chars)


Numpsy:
String.Create is supposed to allocate less than StringBuilder, but I haven't tried testing that here

Numpsy:
Current .NET versions have built in CopyTo functions to copy List<T> and Array<T> into a Span<T>, but I don't think there is anything built in to do the same with FSharpList<T>. Would it be useful to add a local extension method rather than doing the work inline, which could be more reusable elsewhere?

knocte:
Let's see the implementation of that?

knocte (changes requested):
2 things to test/benchmark here instead of current appraoch:

1. Create a CopyTo function to copy List and Array into a Span.
2. Use StringBuilder.

Numpsy:
Ok, I didn't know this before, but apparently in current .NET versions, the overload of ```String.Concat``` which takes an ```IEnumerable<T>``` has specific optimizations for ```IEnumerable<char>```

ref
https://github.com/dotnet/coreclr/pull/14298 and the 'StringConcatCharEnumerable' bit of https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-core-2-1/#string

(The current implementation of which uses the presently internal ValueStringBuilder which uses stack alloced memory for small strings to remove some allocations altogether)

Improvements in the benchmark project here smaller that the work project I tested before, but the change is simpler as well as faster, so:

Before
```
| LintParsedFile | 1.424 s | 0.0112 s | 0.0099 s | 36000.0000 | 12000.0000 | 3000.0000 | 622.11 MB |
```

After
```
| LintParsedFile | 1.413 s | 0.0149 s | 0.0132 s | 36000.0000 | 12000.0000 | 3000.0000 | 622.03 MB |
```
(the numbers are basically the same as the String.Create version, and it's simpler)

knocte:
Looks great now, and code is much simpler this way. Can you just tweak commit msg please:
- Add "Core(HintParser): " scope to the title.
- Make commit msg include all the info about the before/after, no point to leave this info only in the PR.

PR #766, opened 2025-08-18
Makefile: don't skip 'docs' target

webwarrior-ws:
If there is a file or directory named the same as target
('docs' in this case) in current directory, Make treats that
target as a file target. Since the file (dir in this case)
already exists and has no dependencies (or its dependencies
are older), Make considered it "up to date" and skipped execution.
    
To avoid this, we now use a .PHONY target to tell Make that 'docs'
is not a real file.
    
Fixes https://github.com/fsprojects/FSharpLint/issues/768

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -405,7 +405,8 @@ FodyWeavers.xsd
 tmp/
 temp/
 .fsdocs
-docs/
+docs/*
+!docs/Makefile
 
 # ===================================================
 # OS files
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -405,7 +405,8 @@ FodyWeavers.xsd
 tmp/
 temp/
 .fsdocs
-docs/
+docs/*
+!docs/Makefile
```
knocte, line 409: @webwarrior-ws what is this change for?

webwarrior-ws: It works now. Have to run `make` with `--directory` argument.

webwarrior-ws: So abandon 2nd commit and use initial approach?

PR #767, opened 2025-09-08
Update Fornax to 0.16.0

Numpsy:
The FSharp.Core dependency in Fornax.Core has been changed from ```= 8.0.100``` to ```>= 6.0.7``` which should hopefully fix these warnings in the docs build:

image


Numpsy:
commit comment updated

knocte:
I'm actually wondering, shouldn't we turn on Warnings As Errors in order to catch this kind of things earlier?

PR #770, opened 2025-10-19
FSharp.Compiler.Service 43.10.103

Numpsy:
Just having a go at building it with the RC .NET 10 version to see if it all works.

### DIFF
--- a/src/FSharpLint.Core/Framework/Ast.fs
+++ b/src/FSharpLint.Core/Framework/Ast.fs
@@ -259,8 +259,6 @@ module Ast =
             add <| Pattern lhs
             add <| Pattern rhs
 
-
-    // fsharplint:disable FL0025
     let inline private expressionChildren (node: SynExpr) (add: AstNode -> unit) =
         let addMany = List.iter add
 
@@ -280,6 +278,8 @@ module Ast =
         | SynExpr.Lazy(expression, _)
         | SynExpr.TraitCall(_, _, expression, _)
         | SynExpr.YieldOrReturn(_, expression, _, _)
+        | SynExpr.AnonRecd(_, Some (expression, _), _, _, _) ->
+        | SynExpr.IndexFromEnd(expression, _) ->
         | SynExpr.YieldOrReturnFrom(_, expression, _, _) -> add <| Expression expression
         | SynExpr.SequentialOrImplicitYield(_, expression1, expression2, ifNotExpression, _) ->
             addMany [Expression expression1; Expression expression2; Expression ifNotExpression]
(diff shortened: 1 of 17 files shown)


src/FSharpLint.Core/Framework/Ast.fs
```diff
@@ -327,9 +322,22 @@ module Ast =
...
         | SynExpr.Downcast(expression, synType, _) ->
             addMany [Type synType; Expression expression]
-        | SynExpr.LetOrUse(_, _, bindings, expression, _, _) ->
+        // regular let or use
+        | SynExpr.LetOrUse(_, _, _, false, bindings, expression, _, _) ->
             add <| Expression expression
             List.revIter (Binding >> add) bindings
+        // let! or use!
+        | SynExpr.LetOrUse(_, _, _, true, bindings, leftHandSide, _, _) ->
+            match bindings with
```
Numpsy: Change based on the FCS release notes at https://github.com/dotnet/fsharp/blob/c1b13e3d4542d1de7439b34de00384934365aeb9/docs/release-notes/.FSharp.Compiler.Service/10.0.100.md?plain=1#L63

src/FSharpLint.Core/Framework/Ast.fs
```diff
@@ -256,6 +256,8 @@ module Ast =
             add <| Pattern lhs
             add <| Pattern rhs
 
+
+    // fsharplint:disable FL0025
```
Numpsy: temporarily turned this off as the self check was failing because this function is now 105 lines long and selfcheck wants if to be less than 100

knocte:
@Numpsy you're right, our daily CI run has started to fail today. Maybe setup-dotnet doesn't honor `with: global-json-file: global.json` properly anymore? We need to fix/workaround this in case that 674 issue is not fixed.

Numpsy:
It sounds like it's running with the .NET 9 SDK but trying to build as .NET 10 for some reason? (though the error shows the 9.0.306 SDK, and global.json is 9.0.201).
Not sure what's going on there at the moment, I didn't see that error when I built it locally (on Windows)

Numpsy:
I tried changing the CI to actually run as .NET 10 and that fails with a type check error in the build script - https://github.com/Numpsy/FSharpLint/actions/runs/19300021169/job/55193408198#step:7:12

Nothing can ever be simple :-(

Numpsy:
I'll have a look later

Numpsy:
I ran the latest code on a profiler and got this:

image

Which I guess is coming from https://github.com/webwarrior-ws/FSharpLint/blob/c378757ad49d9a0c9eedbf228ced722245643978/src/FSharpLint.Core/Rules/Conventions/Naming/AsynchronousFunctionNames.fs#L36 which was recently added?

With the current FCS9 it gets this, which is still a hotspot in the same place, but doesn't take nearly as much time

image

I've seen a number of bugs raised in the F# repo about this cache stuff causing performance issues of which some are fixed and some which aren't, though I'm sure I've read of issues elsewhere related to getting those project contexts lots of times but I can't recall the details.

Numpsy:
fwiw, I queried the perf diference between the FCS versions at https://github.com/dotnet/fsharp/discussions/19294  in case there's a bug in it

Numpsy:
Ok, completing #845 has fixed the performance degredation previously caused by the FCS update.

knocte:
Superseded by https://github.com/fsprojects/FSharpLint/pull/850

PR #777, opened 2025-11-10
Wildcard support added

### LINKED ISSUE #776: Support wildcards
When picking which files to analyse, I would like to do `*.fsx` instead of only specific `fs`-file or `fsproj`-file or `sln`-file. This should support recursion over subdirectories.

Thorium:
Fixes #776

### DIFF
--- a/src/FSharpLint.Console/Program.fs
+++ b/src/FSharpLint.Console/Program.fs
@@ -63,20 +63,20 @@ let internal expandWildcard (pattern:string) =
         | -1 ->
             // Non-recursive pattern
             match normalizedPattern.LastIndexOf '/' with
-            | -1 -> ".", normalizedPattern, SearchOption.TopDirectoryOnly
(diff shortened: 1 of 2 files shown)


knocte:
Thanks @Thorium good work! Just from a quick glance, can you:

1. Fix CI build please. It's again failing in the SelfCheck step, those issue are easy to address IMO, ping me if you need help.
2. Can you add some kind of warning when using this command line option? Because it's still recommended/preferable that users use their solutions or projects, rather than individual files, because this way FSharpLint can act on the typed tree (and discover more warnings than if it were running on an untyped one).

Thorium:
Will check. The 2 major use-cases I have:
1) I have a group of fsx files in random directory structure to check. There is no project.
2) I have installed .NET10 runtime to a computer and global.json missing from a project, so FSharpLint fails for System.Runtime v10 errors.

Thorium:
I think the warning is still ok to show: The best practice is not work with big bunch of fsx files because the compiled F# is faster.

PR #790, opened 2025-12-04
GithubCI: split "deployReleaseBinaries" job in two

webwarrior-ws:
Split the job deployReleaseBinaries in two:
- One called packReleaseBinaries, that just uploads stuff as artifact.
- Another one called publishReleaseBinaries, that uploads nuget packages to nuget.org.
- The 2 integration tests run after packReleaseBinaries is green.
- After the 2 integration tests pass, then publishReleaseBinaries runs (using packReleaseBinaries' artifacts, with the download-artifact action).

### DIFF
--- a/.github/workflows/build+test+deploy.yml
+++ b/.github/workflows/build+test+deploy.yml
@@ -37,7 +37,7 @@ jobs:
       run: make selfcheck
 
 
-  deployReleaseBinaries:
+  packReleaseBinaries:
     needs: buildAndTest
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build+test+deploy.yml
```diff
@@ -54,6 +54,21 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Restore tools
+      run: dotnet tool restore
+    - name: Download artifact
+      uses: actions/download-artifact@v4
+      with:
+        name: binaries
+        path: ./out/
+    - run: ls -l ./out/
```
knocte: @webwarrior-ws i guess this line is just for testing, can remove now?

webwarrior-ws: Removed.

Issue #792, opened 2025-12-09
FavourStaticEmptyFields false negative


knocte:
In FSharpLint code itself: TrailingNewLineInFile.fs L14, ...

webwarrior-ws:
https://github.com/webwarrior-ws/FSharpLint/commits/refs/heads/fix-792/

But sometimes the rule fires twice for the same expression.

PR #798, opened 2025-12-23
Fix FavourStaticEmptyFields false negative

webwarrior-ws:
* Fixed false-negatives in FavourStaticEmptyFields rule.
* Added 3 test cases for this rule.
* Fixed FSharpLint SelfCheck violations due to rule working properly now.

Fixes https://github.com/fsprojects/FSharpLint/issues/792

### DIFF
--- a/src/FSharpLint.Core/Rules/Conventions/FavourStaticEmptyFields.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/FavourStaticEmptyFields.fs
@@ -47,7 +47,15 @@ let rec private processExpressions (errorsSoFar: array<WarningDetails>) (args: A
                 errorsSoFar
                 (generateError args.FileContent range EmptyStringLiteral)
         processExpressions errors args tail
-    | SynExpr.ArrayOrList (isArray, [], range) :: tail ->
+    | SynExpr.DotIndexedGet(_, indexArgs, _ , _) :: tail ->
+        processExpressions errorsSoFar args (indexArgs :: tail)
+    | SynExpr.DotIndexedSet(_, indexArgs, _ , _, _, _) :: tail ->
+        processExpressions errorsSoFar args (indexArgs :: tail)
+    | SynExpr.Paren(expr, _, _, _) :: tail ->
+        processExpressions errorsSoFar args (expr :: tail)
+    | SynExpr.Tuple(_, expressions, _, _) :: tail ->
+        processExpressions errorsSoFar args (List.append expressions tail)
+    | SynExpr.ArrayOrList(isArray, [], range) :: tail ->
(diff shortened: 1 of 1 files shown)


tests/FSharpLint.Core.Tests/Rules/Conventions/FavourStaticEmptyFields.fs
```diff
@@ -118,6 +118,39 @@ type Person =
...
 
+    [<Test>]
+    member this.FavourStaticEmptyFieldsShouldProduceError13() =
+        this.Parse """
+let foo =
+    {
+        Range = Range.mkRange "" pos pos
+        Message = Resources.GetString("RulesTypographyTrailingLineError")
+        SuggestedFix = None
+        TypeChecks = List.Empty
```
knocte: @webwarrior-ws let's remove unnecessary record elements to simplify the testcase

webwarrior-ws:
> 1st commit of this PR is a refactoring before you add any failing tests, so it should have green CI.

It is not green because the function that I introduce is not tail-recursive.

I fix it in "FavourStaticEmptyFields: made function tail-recursive" commit.

knocte:
> Then I'll need to squash all commits that change the rule code.

No need because those have different CI errors that are being fixed in an tiered way.

webwarrior-ws:
> I disagree, you only have to squash 2 commits here: [ee608bf](https://github.com/fsprojects/FSharpLint/commit/ee608bf99ef659336e5c0b652f69e737bbcc8d8e) with [cd76ae6](https://github.com/fsprojects/FSharpLint/commit/cd76ae6c0ec9dcd84b7b5da3cbe48f5a41d2bcdb)

But I can't just squash these 2. Because there were changes to the function in another 2 commits in-between those.
And making the function tail-recursive is not just adding an attribute, but rewriting it (you can see in [cd76ae6](https://github.com/fsprojects/FSharpLint/commit/cd76ae6c0ec9dcd84b7b5da3cbe48f5a41d2bcdb) that almost all lines are changed).

knocte:
Then you gotta squash manually. Refactorings should never break CI.

knocte:
I don't understand how 4fbb1cb can now have red CI.

webwarrior-ws:
> > I don't understand how [4fbb1cb](https://github.com/fsprojects/FSharpLint/commit/4fbb1cb5783b829c01f7053a647f419920a1af29) can now have red CI.
> 
> Actually, that means that it's a bad squash, a refactoring should not change behaviour.

In this case, no. Reason for CI failure is self-check violation in `src/FSharpLint.Core/Rules/Hints/HintMatcher.fs`.

PR #808, opened 2025-12-29
AvoidSinglePipeOperator: allow single pipe operator with multiline argument

webwarrior-ws:
Do not apply this rule when an argument passed to pipe operator spans multiple lines.

### DIFF
--- a/src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
@@ -8,6 +8,9 @@ open FSharp.Compiler.Syntax
 open FSharpLint.Framework.Ast
 open FSharpLint.Framework.Rules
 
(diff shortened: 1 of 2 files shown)


src/FSharpLint.Core/Rules/Conventions/AvoidSinglePipeOperator.fs
```diff
@@ -8,6 +8,9 @@ open FSharp.Compiler.Syntax
 open FSharpLint.Framework.Ast
 open FSharpLint.Framework.Rules
 
+let isSingleLine (range: FSharp.Compiler.Text.range) =
+    range.EndLine  = range.StartLine
```
knocte: @webwarrior-ws please squash 2 spaces into 1

webwarrior-ws:
> * Please squash 2nd with 3rd commit, no reason to separate them.
> * Please avoid ending dots in commit msg titles.

Squashed last 2 commits and fixed commit message title.

Issue #545, opened 2022-05-02
Rule to recommend if-else construct instead of match clause


knocte:
### Description

If a match clause only has 2 cases, and the last case is _, then the whole thing can be converted into a simple if-then-else block, which is more readable.

Test1:
```
match foo with
| bar -> ()
| _ -> ()
```
Should throw warnings.

Test2:
```
match foo with
| bar -> ()
| baz -> ()
| _ -> ()
```
Should not throw warnings (because it has 3 cases, not 2).

Test3:
```
match foo with
| bar -> ()
| baz -> ()
```
Should not throw warnings (because its last case is not _).

Test4:
```
match foo with
| Bar baz -> ()
| _ -> ()
```
Should not throw warnings (because its first case is not a single element, so the casting is useful).

Thorium:
I think match is very readable if you have used to it.
How would you made this more readable by if clause?

```fsharp
match a when
| Some x when x > 21 -> x
| _ -> 21
```

PR #588, opened 2022-11-21
Rule to recommend if-else construct instead of match clause

parhamsaremi:
It improves code readability.

Fixes: https://github.com/fsprojects/FSharpLint/issues/545

**NOTE**: this PR supersedes an old [PR](https://github.com/fsprojects/FSharpLint/pull/546)

PR #822, opened 2026-01-20
Rule to recommend if-else construct instead of match clause

webwarrior-ws:
It improves code readability.

Fixes: https://github.com/fsprojects/FSharpLint/issues/545

Supersedes https://github.com/fsprojects/FSharpLint/pull/588

### DIFF
--- a/src/FSharpLint.Core/Rules/Conventions/RecommendIfElseConstructOverMatch.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/RecommendIfElseConstructOverMatch.fs
@@ -9,20 +9,18 @@ open FSharpLint.Framework.Rules
 
 let runner args =
     let isBasicControlFlow (synMatchClauses: List<SynMatchClause>) =
-        let isFirstClauseTarget firstClause =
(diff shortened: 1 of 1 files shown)


knocte:
From https://github.com/fsprojects/FSharpLint/pull/586:
> btw what do you think about renaming the rule to RecommendIfElseConstructOverMatch?

tests/FSharpLint.Core.Tests/Rules/Conventions/RecommendIfElseConstructOverMatch.fs
```diff
@@ -34,6 +34,23 @@ match foo with
...
+
+    [<Test>]
+    member this.ThePresenceOfNullaryDUCaseShouldNotProduceError() =
+        this.Parse """
+type Baz =
+    | Bar
+
+match foo with
+| Bar -> ()
+| _ -> ()
```
knocte: @webwarrior-ws give me an example of a type that `foo` could have that would fall under this `_` case?

PR #848, opened 2026-03-03
Restore the console project before running a self-lint on it.

Numpsy:
This also requires updating the test nuget.config file so that nuget.org is reachable. It now uses package mapping to force the usage of a local fsharplint package.

refs the current CI build failiures in the 'testReleaseBinariesWithDotNet10' step. I don't know why the nightly build started failing when the source hadn't changed, but restoring the console project explicitly before doing the lint does appear to fix the current 'Lint failed to parse files' errors.

Maybe this is a simple way to get the CI working again at least.

### DIFF
--- a/.github/workflows/build+test+deploy.yml
+++ b/.github/workflows/build+test+deploy.yml
@@ -187,16 +187,25 @@ jobs:
           <?xml version="1.0" encoding="utf-8"?>
           <configuration>
             <packageSources>
+              <clear />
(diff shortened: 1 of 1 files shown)


Numpsy:
The CI appears to have started working again after the March monthly patch updates, so maybe it was a bug in the dotnet SDK which is fixed now (just a guess, given that it broke at the end of february and started working again at the end of March).

Will close this for now anyway

PR #850, opened 2026-03-15
FSharp.Compiler.Service 43.12.201 

Numpsy:
This contains the whole of https://github.com/fsprojects/FSharpLint/pull/770, plus an extra commit to update to FCS  43.12.201 (I'm not clear what's going on with the version numbers here)

Needs a review to see if the 'ExpressionUtilities.LetOrUse' helper is the right approach.

### DIFF
--- a/src/FSharpLint.Core/Framework/Utilities.fs
+++ b/src/FSharpLint.Core/Framework/Utilities.fs
@@ -153,6 +153,15 @@ module ExpressionUtilities =
         (range.StartLine, range.StartColumn) >= (containingRange.StartLine, containingRange.StartColumn) 
             && (range.EndLine, range.EndColumn) <= (containingRange.EndLine, containingRange.EndColumn)
 
+    /// Active pattern to match any SynExpr.LetOrUse
+    /// Returns a tuple of (record, isBang, isUse) allowing matching on both booleans and accessing the full record
+    /// Borrowed from https://github.com/nojaf/fsharp/blob/b3d90dc6ab9a9cedad4b8702fd8625f8f8175ae1/src/Compiler/SyntaxTree/SyntaxTreeOps.fs#L138
+    [<return: Struct>]
+    let (|LetOrUse|_|) (expr: SynExpr) =
+        match expr with
+        | SynExpr.LetOrUse(letOrUse) -> ValueSome(letOrUse, letOrUse.IsBang, letOrUse.IsUse)
+        | _ -> ValueNone
+
 module String =
 
     open System.IO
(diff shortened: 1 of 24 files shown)


src/FSharpLint.Core/Rules/Conventions/FavourNonMutablePropertyInitialization.fs
```diff
@@ -64,17 +64,17 @@ let rec private processLetBinding (instanceNames: Set<string>) (body: SynExpr) (
 and [<TailCall>] processExpression (expression: SynExpr) (continuation: unit -> array<WarningDetails>) : array<WarningDetails> =
     Array.append
         (match expression with
-        | SynExpr.LetOrUse(_, _, bindings, body, _, _) ->
-            let instanceNames = extraFromBindings bindings List.Empty |> Set.ofList
-            processLetBinding instanceNames body returnEmptyArray
+        | SynExpr.LetOrUse letOrUse when not letOrUse.IsBang ->
```
Numpsy, line 67: This one is using ```when``` because when I tried the active pattern I got ```error FS3569: The member or function 'processLetBinding' has the 'TailCallAttribute' attribute, but is not being used in a tail recursive way.```

Numpsy:
Looks like the performance is still bad without https://github.com/fsprojects/FSharpLint/pull/845

src/FSharpLint.Core/Rules/Conventions/FavourNonMutablePropertyInitialization.fs
knocte: That's what the TailCall attrib is for: so that the compiler makes sure you're using a tail call recursive algorithm. If you mark a func as TailCall without changing the algorithm to be a tail-call-recursive, compilation will fail.

knocte:
Ok, this version of FCS should contain the fix we need (https://github.com/dotnet/fsharp/commit/89d788641914c5d0b87fddfa11f4df0b5cfaa73d), can you please include a commit to remove this workaround: 755f4d1d6eaf3dcfadaac9528a04c1179aea9492 ?

Numpsy:
> Ok, this version of FCS should contain the fix we need ([dotnet/fsharp@89d7886](https://github.com/dotnet/fsharp/commit/89d788641914c5d0b87fddfa11f4df0b5cfaa73d)), can you please include a commit to remove this workaround: [755f4d1](https://github.com/fsprojects/FSharpLint/commit/755f4d1d6eaf3dcfadaac9528a04c1179aea9492) ?

Done

src/FSharpLint.Core/Framework/Utilities.fs
```diff
@@ -153,6 +153,15 @@ module ExpressionUtilities =
         (range.StartLine, range.StartColumn) >= (containingRange.StartLine, containingRange.StartColumn) 
             && (range.EndLine, range.EndColumn) <= (containingRange.EndLine, containingRange.EndColumn)
 
+    /// Active pattern to match any SynExpr.LetOrUse
+    /// Returns a tuple of (isBang, isUse, record) allowing matching on both booleans and accessing the full record
```
webwarrior-ws: Comment does not match what the function returns. It should be `(record, isBang, isUse)`

webwarrior-ws (commented):
Otherwise LGTM.

PR #852, opened 2026-04-26
Change String::toLines to use ValueOption instead of Option

Numpsy:
If reduces memory allocations slightly

Just a thought.

Current benchmark project allocation results
```
| Method         | Mean     | Error   | StdDev  | Gen0       | Gen1      | Gen2      | Allocated |
|--------------- |---------:|--------:|--------:|-----------:|----------:|----------:|----------:|
| LintParsedFile | 842.2 ms | 5.43 ms | 4.81 ms | 25000.0000 | 8000.0000 | 1000.0000 | 397.55 MB |
```

With this change
```
| Method         | Mean     | Error   | StdDev  | Gen0       | Gen1      | Gen2      | Allocated |
|--------------- |---------:|--------:|--------:|-----------:|----------:|----------:|----------:|
| LintParsedFile | 837.7 ms | 4.83 ms | 4.29 ms | 25000.0000 | 8000.0000 | 1000.0000 | 396.83 MB |
```

### DIFF
--- a/src/FSharpLint.Core/Framework/Utilities.fs
+++ b/src/FSharpLint.Core/Framework/Utilities.fs
@@ -150,16 +150,16 @@ module String =
     open System.IO
 
     [<TailCall>]
-    let rec private iterateLines (lines: ResizeArray<string*int*bool>) (readLine: unit -> Option<string>) currentLine index =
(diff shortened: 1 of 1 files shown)


knocte:
Thanks @Numpsy and sorry for the delay.

For next PRs please try to make git commit msg title as short as possible, while you can make the commit msg body as long as possible (even matching with the long explanation in the PR description).

PR #857, opened 2026-06-04
Fix issue 856

### LINKED ISSUE #856: bug: Interpolated strings with different literal text content wrongly considered equal
## Description

The built-in hint `if x then y else y ===> y` triggers on `if/then/else` expressions where both branches are interpolated strings (`$"..."`) that contain the same fill expressions but **different literal text segments**. As the literal text segments are different, the branches should not be considered equivalent.

## Repro

`fsharplint.json` (relevant excerpt — using the default hints shipped with FSharpLint):

```json
{
  "hints": {
    "add": [
      "if x then y else y ===> y"
    ]
  }
}
```

Source:

```fsharp
let label (p: int) (currentPage: int) =
    if p = currentPage then $"Page {p}" else $"Goto page {p}"
```

## Expected

No warning. The two branches have different interpolated sources (`"Page 3"` vs `"Goto page 3"`).

## Actual

FSharpLint reports the `if x then y else y ===> y` hint and suggests replacing the whole expression with one of the branches, which would change behavior.

## Notes

- Making the fill expressions different from each other, e.g. `if p = currentPage then $"Page {p}" else $"Goto page {currentPage}"` suppresses the warning, suggesting only the fill expressions are compared for equality, ignoring the surrounding text
- Replacing the interpolated strings with `sprintf` (`sprintf "Page %d" p` / `sprintf "Goto page %d" p`) suppresses the warning, as a workaround
- Reproduced with FSharpLint version: `0.26.10`
- .NET SDK: `10.0.108`

webwarrior-ws:
Made getHashCode take into account all parts of interpolated strings. This is needed for hint matcher to distinguish between interpolated strings with the same fill expressions but different literal text segments.

Also added a regression test.

Fixes https://github.com/fsprojects/FSharpLint/issues/856

### DIFF
--- a/src/FSharpLint.Core/Framework/AbstractSyntaxArray.fs
+++ b/src/FSharpLint.Core/Framework/AbstractSyntaxArray.fs
@@ -202,6 +202,9 @@ module AbstractSyntaxArray =
         | Expression(SynExpr.Const(SynConst.UInt64(x), _)) -> hash x
         | Pattern(SynPat.Const(SynConst.UIntPtr(x), _))
         | Expression(SynExpr.Const(SynConst.UIntPtr(x), _)) -> hash x
+        | Expression(SynExpr.InterpolatedString(contents, _, _)) ->
+            contents 
+            |> List.fold (fun acc part -> acc ^^^ (part.ToString() |> hash)) 0
         | _ -> 0
 
     [<Struct; NoEquality; NoComparison>]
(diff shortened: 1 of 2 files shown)


knocte:
@webwarrior-ws please push 1 by 1

webwarrior-ws:
> @webwarrior-ws please push 1 by 1

Pushed.

PR #846, opened 2026-02-24
Performance: parallel per-file type checking

michaelglass:
## Summary

Massive performance improvement to the linting pipeline, built on top of #845.

Runs `ParseAndCheckProject` and per-file `ParseAndCheckFileInProject` concurrently using `Async.Parallel`. The project check warms the FCS incremental builder; parallel per-file checks share the builder and benefit from its warmed state.

Also enables `parallelReferenceResolution` on `FSharpChecker.Create`.

### Performance

| Project | Before (sequential) | After (parallel) | Speedup |
|---|---|---|---|
| FSharpLint.Core (15 files) | 15.4s | 12.5s | ~1.2x |
| my internal project (48 files, many huge reps) | 658s | ~25s | ~25x |

The speedup scales with file count — larger projects see dramatically larger gains.

## Test plan

- [x] Full test suite passes (1003 unit + 26 console + 13 functional)
- [x] Verified identical warning count between sequential and parallel approaches
- [x] Type-check-dependent rules (FL0065 hints, FL0085 tail call diagnostics, FL0095 sync naming) produce correct results in parallel mode

## A note on this PR

This PR was developed with significant AI assistance. I understand if that's not welcome — apologies if so. Happy to discuss the changes or withdraw the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #861, opened 2026-06-29
Parallel per-file type checking with concurrent project check

michaelglass:
sorry, force pushed so need to open a new PR for #846 

this _does_ help for projects with lots of local dependencies. E.g. fsharp monorepos with multiple dependent projects, for instance https://github.com/michaelglass/fshotwatch

[here's my [claude's] benchmark script](https://gist.github.com/michaelglass/53cbde5b806345d1ede470547854278a)

results:

pathological target -- linting FSHotWatch
| stage | solution lint | vs master |
|---|---:|---:|
| master (`91fd00b0`, incl. #845) | ~908 s | 1× |
| + #846 parallel per-file (variant C, `ee218763`) | ~91 s | ~10× |
| + cross-project sharing (`78f134b3`) | **~27 s** | **~33×** |
_0.27.0 (released) ≈ 1000 s for reference. Sharing adds **2.28×** on top of #846._

Non-pathological target — linting FSharpLint's own solution (flat dependency graph):

| linter | FSharpLint.slnx lint | vs master |
|---|---:|---:|
| `0.27` (released) | ~27 s | ~1.0× |
| master (`91fd00b0`) | ~27 s | 1× |
| + #846 parallel per-file (`ee218763`) | ~27 s | ~1.0× (no regression) |
| + cross-project sharing (`78f134b3`) | ~28 s | ~0.96× (≈ +1 s / ~2–4%) |

**What makes this pathological / what the speedup is conditional on**

The perf improvement scales with a solution of **many local subprojects that reference each other**. FsHotWatch is a good example: 13 projects, with the test/CLI projects each referencing 8–9 of the others.

On master, every project is linted in isolation — a fresh MSBuild design-time load + FCS re-type-checking that project's whole project-to-project (P2P) dependency closure. So each shared library project gets re-loaded and re-type-checked **once per dependent**: duplicated work that grows with the density of the dependency graph.

The two changes remove different parts of that cost:
- **#846 parallel per-file (~10×):** type-checks a project's files in parallel instead of serially — conditional on projects having **many files**.
- **Cross-project sharing (→~33×):** one `WorkspaceLoader` (MSBuild design-time build runs **once for the whole solution**, not per project) + one shared `FSharpChecker` (each project and its dependencies type-checked **once**, cached, reused by every dependent) — conditional on **many cross-referencing local subprojects**.

**the ~33× is conditional on lots of inter-dependent local subprojects.** 

## Caveat

On a single project, or a flat solution whose projects share no dependencies, the sharing win largely disappears and you're left with roughly the parallel-per-file effect; on a single small project the parallelization overhead can even make it marginally slower.

### DIFF
--- a/src/FSharpLint.Console/Program.fs
+++ b/src/FSharpLint.Console/Program.fs
@@ -157,6 +157,7 @@ let main argv =
                 ReceivedWarning = Some output.WriteWarning
                 Configuration = configParam
                 ReportLinterProgress = Some (parserProgress output)
+                Checker = None
             }
 
         let target = lintArgs.GetResult Target
--- a/src/FSharpLint.Core/Application/Lint.fs
+++ b/src/FSharpLint.Core/Application/Lint.fs
@@ -436,9 +436,12 @@ module Lint =
 
                     Option.iter (fun func -> func warning) optionalParams.ReceivedWarning
 
+                // parallelReferenceResolution=true was measured here and intentionally left off:
+                // on FCS 43.9.201 it gave no wall-clock improvement and ~20% more CPU
+                // (FsHotWatch.slnx: ~56s/74 CPU-s without it vs ~58s/94 CPU-s with it).
                 let checker = FSharpChecker.Create(keepAssemblyContents=true)
 
-                let parseFilesInProject files projectOptions = async {
+                let parseFilesInProject files (projectOptions:FSharpProjectOptions) = async {
                     let lintInformation =
                         { Configuration = config
                           CancellationToken = optionalParams.CancellationToken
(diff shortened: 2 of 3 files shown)


knocte:
@michaelglass hey thanks, do you mind fixing CI? But please fix it so that each commit has a green CI status.

michaelglass:
well https://github.com/michaelglass/FSharpLint/commit/7fc3684882947353ea0aa09f2165261c4ba97344 is green, [but not in this PR?](https://github.com/fsprojects/FSharpLint/pull/861/changes/7fc3684882947353ea0aa09f2165261c4ba97344) 🤷

michaelglass:
I'm going to also benchmark this against non-pathological case to see how it performs as well. I know when I did this initially there was a minor regression (single-digit percentage)

knocte:
It's not that it's not green in the PR, it's that in the PR it's missing CI status because you didn't push 1 by 1 :) I made a script for that:  https://github.com/tarsgate/conventions/blob/master/scripts/gitPush1by1.fsx . But don't worry, I know it's a hassle, I can take care of it this weekend.

michaelglass:
ok can go back and push one by one again. no stress

knocte:
I stand corrected, what must be actually happening is that, despite you having pushed 1 by 1, as you're a first time contributor, I need to click "approval to run CI" everytime you push. Hopefully this doesn't happen anymore after your first PR lands.

michaelglass:
ok done. no serious regression!

michaelglass:
added the benchmark scrip to [the gist](https://gist.github.com/michaelglass/53cbde5b806345d1ede470547854278a)

Numpsy:
Out of interest, did the change in https://github.com/fsprojects/FSharpLint/pull/884 improve the 'pathological target' numbers? It made a big different in the benchmark project, but I don't have any figures for the different on a large solution

michaelglass:
@numpsy nope. Here are the new benchmark results

whole solution (FsHotWatch, 13 projects):

| | mean |
|---|---:|
| `beed0e5a` (just before #884) | ~1165 s |
| `d2607f5c` (#884) | ~1177 s |

and one project, 10 runs each:

| | before | after |
|---|---:|---:|
| `src/FsHotWatch` | 9.91 ± 0.69 s | 9.87 ± 0.99 s |
| `src/FsHotWatch.TestPrune` | 10.13 ± 0.41 s | 9.97 ± 0.51 s |

looks about the same — the gap is smaller than the run-to-run noise.

#884 makes **running the rules** faster (scanning a file's text to pull out the code a rule is looking at). but my pathological case is dominated by duplicate **type-checking**: every project gets linted on its own, so each shared library gets re-type-checked once per dependent.

PR #864, opened 2026-07-06
NoPartialFunctions: fixed bug on object expressions

webwarrior-ws:
By removing incorrect type cast.

Also added a regression test.

### DIFF
--- a/src/FSharpLint.Core/Rules/Conventions/NoPartialFunctions.fs
+++ b/src/FSharpLint.Core/Rules/Conventions/NoPartialFunctions.fs
@@ -173,7 +173,7 @@ let rec private tryFindTypedExpression (range: Range) (expressions: List<FSharpE
             |> List.collect (fun (_, imlps) -> imlps)
         let exprs =
             List.append overrides interfaceImlps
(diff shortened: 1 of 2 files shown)


knocte:
@webwarrior-ws just realised, Seq.cast is a partial function! So NoPartialFunctions should have detected this at SelfCheck, please cook a PR that catches Seq.cast and suggests Seq.choose because Seq.tryCast doesn't exist yet.

E.g.:
```
// 1. Define the tryCast helper function
let tryCast<'T> (item: obj) =
    match item with
    | :? 'T as casted -> Some casted
    | _ -> None

// 2. Use it with Seq.choose to filter and cast elements
let mixedCollection = [ box "Hello"; box 42; box "World"; box 3.14 ]

let stringsOnly =
    mixedCollection
    | > Seq.choose tryCast<string>

// stringsOnly now contains: seq ["Hello"; "World"]
```

Then after that cook a PR for https://github.com/fsprojects/FSharpx.Collections to add Seq.tryCast there, and when that gets merged and released, we can change the recommendation in FSharpLint from Seq.choose to FSharpx.Collections.Seq.tryCast

PR #865, opened 2026-07-07
NoPartialFunctions: add Seq.cast to list of partial function identifiers

webwarrior-ws:
`Seq.cast` is a partial function because it will raise an exception at runtime if one of the elements can't be cast to specified type.
`Seq.choose tryUnbox` is proposed as an alternative.


knocte:
@webwarrior-ws next time, push the commit of the failing test first, then create PR, then push one by one; otherwise all CI status won't appear here.

PR #866, opened 2026-07-08
Fix hints bug when different records are treated as equal

webwarrior-ws:
Made `getHashCode` take into account inner expressions of record expressions. This is needed because record is treated as a single leaf node in FSharpLint AST and hint matcher compares only hashes of ast nodes.

Function `getHashCode` was refactored to make it tail-recursive, as now it has to analyze inner expressions.

### DIFF
--- a/src/FSharpLint.Core/Framework/AbstractSyntaxArray.fs
+++ b/src/FSharpLint.Core/Framework/AbstractSyntaxArray.fs
@@ -163,49 +163,68 @@ module AbstractSyntaxArray =
         | [] -> 0
 
     /// Get hash code of an ast node to be used for the fuzzy match of hints against the ast.
-    let private getHashCode node =
-        match node with
-        | Identifier(idents, _) -> getIdentHash idents
-        | Pattern(SynPat.Const(SynConst.Bool(x), _))
-        | Expression(SynExpr.Const(SynConst.Bool(x), _)) -> hash x
-        | Pattern(SynPat.Const(SynConst.Byte(x), _))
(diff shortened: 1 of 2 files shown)


src/FSharpLint.Core/Framework/AbstractSyntaxArray.fs
```diff
@@ -163,49 +163,68 @@ module AbstractSyntaxArray =
...
-        | Expression(SynExpr.Const(SynConst.UInt64(x), _)) -> hash x
-        | Pattern(SynPat.Const(SynConst.UIntPtr(x), _))
-        | Expression(SynExpr.Const(SynConst.UIntPtr(x), _)) -> hash x
-        | Expression(SynExpr.InterpolatedString(contents, _, _)) ->
-            contents 
-            |> List.fold (fun acc part -> acc ^^^ (part.ToString() |> hash)) 0
-        | _ -> 0
+    [<TailCall>]
+    let rec private getHashCode (acc: int) (nodes: list<AstNode>) =
+        match nodes with
```
knocte, line 168: @webwarrior-ws please start the func with a `let` statement that will save the result of the `match`, otherwise using the `^^^` immediately afterwards looks weird

webwarrior-ws: If I do that, compiler will complain that the function is not used in tail-recursive way.

knocte: are you sure?

webwarrior-ws: Yes, I tried that.

PR #868, opened 2026-07-10
tests(Benchmarks): Enable allocation reporting in the lint benchmark

Numpsy:
This allows the benchmark project to be used to test changes intended to reduce memory allocations


Numpsy:
I've had this enabled locally before for testing things like https://github.com/fsprojects/FSharpLint/pull/765, but I think it'd be useful to have it enabled by default

knocte:
@Numpsy please add a scope to the commit msg title: `tests(Benchmarks): `

Issue #869, opened 2026-07-12
Possible issue with SourceLengthHelper::stripMultilineComments ?


Numpsy:
Hi,

I was having a look at the code around https://github.com/fsprojects/FSharpLint/blob/bd94a097d9cbb48604c8657b8fa108e7cfd8bf12/src/FSharpLint.Core/Rules/Conventions/SourceLength/SourceLengthHelper.fs#L63 as it seems quite allocation heavy, and when looking noticed that it didn't *appear* to behave correctly when there are multiple block comments present, as it removes the blocks starting from the start of the source string, and each removal changes the offsets of any following comments.

Basically, If I add a new test like
```
    [<Test>]
    member this.FunctionTooManyLinesWithMultiLineComments() =
        this.Parse($"""
module Program

let dog x =
    (*
    Foo
    Bar
    *)
    (*
    Foo
    Bar
    *)
    %s{generateNewLines (FunctionLength - 8) 4}
    ()""")
        Assert.IsFalse this.ErrorsExist
```

Then the generated source is

```
dog x =
    (*
    Foo
    Bar
    *)
    (*
    Foo
    Bar
    *)
    printf System.String.Empty
    printf System.String.Empty
    printf System.String.Empty
    printf System.String.Empty
    printf System.String.Empty

...
```

And after ```stripMultilineComments``` has run it looks like
```
dog x =
    
    (*
    Foo
    Bar
    *).Empty
    printf System.String.Empty
    printf System.String.Empty
    printf System.String.Empty
    printf System.String.Empty

...
```

Which doesn't seem right?
Basically I wonder if it should maybe be removing the blocks in reverse order (starting from the one latest in the file), so the offsets for earlier comments don't change?

knocte:
> Which doesn't seem right?

It is not right indeed. I guess this didn't cause major issues because it is just for measuring source code length (so that output is not used for anything elese other than counting). Let's add a unit test for stripMultilineComments and fix it?

Numpsy:
As it stands, ```stripMultilineComments``` is nested inside ```checkSourceLengthRule``` so I don't think it can be tested directly without moving it.

Additional test observation: If I have a test where there are comments near the end of the source, for example

https://github.com/Numpsy/FSharpLint/commit/420d35722e58a9f84fa2a755156079cc07fb2840 

then the whole thing blows up with an ```ArgumentOutOfRangeException```, so the issue is a little more than just the line counts being potentially wrong:
```
[...]

PR #870, opened 2026-07-12
perf(Core/SourceLength): Change Substring to Span

Numpsy:
I was looking at this when I opened https://github.com/fsprojects/FSharpLint/issues/869 - leaving it as a draft for now due to that
--------------

Creating two substrings and then joining them allocates 3 strings, but creating two spans and concatenating those only allocates one, which can be a substantial saving on large files.

Running the benchmarks with the extended set of 'self check' rules enabled gives me -

Before:
| LintParsedFile | 1.467 s | 0.0287 s | 0.0282 s | 36000.0000 | 12000.0000 | 3000.0000 |    622 MB |

After:
| LintParsedFile | 1.433 s | 0.0085 s | 0.0076 s | 35000.0000 | 11000.0000 | 2000.0000 | 588.75 MB |


knocte:
LGTM; just respect the 50char max limit for commit msg title (otherwise it overflows into next line), e.g. `perf(Core/SourceLength): spans over substrings`, thanks

PR #871, opened 2026-07-14
Remove conditionless reduction hint

webwarrior-ws:
Removed conditionless reduction ("if x then y else y ===> y") hint because it caused false positives on records.

Fixing it would take a lot of work since AbstractSyntaxArray.getHashCode function would have to be defined for all expression types (there are about 60 of them).

A test with `[<Ignore>]` attribute was added as an example of such false positive.

### DIFF
--- a/src/FSharpLint.Core/fsharplint.json
+++ b/src/FSharpLint.Core/fsharplint.json
@@ -431,7 +431,6 @@
             "false <> a ===> a",
             "if a then true else false ===> a",
             "if a then false else true ===> not a",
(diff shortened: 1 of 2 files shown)


knocte:
When you perform a 'git revert' command, git automatically injects the commit hash in the commit msg, why did you remove the commit hash from it?

webwarrior-ws:
> When you perform a 'git revert' command, git automatically injects the commit hash in the commit msg, why did you remove the commit hash from it?

Added commit hash back to the commit message.

